PR #4, opened 2021-08-23
Formatting, types, tidy ups

chidg:
WIP, nothing urgent.


src/components/Panel.tsx
```diff
@@ -1,13 +1,20 @@
...
 
-const PanelSection = styled(({ title, children, className, ...rest }) => {
+type BasePanelProps = {
+  title: string
+  subtitle: string
+  className: string
+  rest: any[]
+}
+
+export const PanelSection = styled(({ title, children, className }) => {
```
chidg, line 10: @alecdwm @danielharrison-flex appreciate your thoughts on this change. TS was not able to understand the pattern where `Section` is an attribute of `Panel`, so I broke them up.

daniel-flex:
@chidg yeah had this problem too, it's a pain as I like that 'dot-notation' pattern for children. I think I've solved it in the extension code in a slightly better way, maybe?

jonathanpdunne (approved):

PR #93, opened 2021-11-03
refactor: hardcode coingecko coins list for now

juliuscarvajal:


juliuscarvajal:
Turns out it's a `gql` issue.

PR #115, opened 2021-11-06
added crowdloan banner

daniel-flex:


daniel-flex:
Added temporary Crowdloan notification for DOT

PR #118, opened 2021-11-07
updated lit-per--dot

daniel-flex:


daniel-flex:
updated lit-per-dot number

PR #180, opened 2021-12-07
remove subdao from _config

jonathanpdunne:


jonathanpdunne:
- SubDAO have ended their intention to secure a parachain slot in the short term:

https://subdao.medium.com/reconsideration-on-polkadot-parachain-slot-auction-7beaefe43495

PR #210, opened 2022-01-20
Updated text from 'benefits' to 'Where can I get one?' 

chidg:
- Updated text from 'benefits' to 'Where can I get one?' for spirit key page.
- Bonus spanish translation

TODO: Chinese translation


juliuscarvajal (approved):
LGTM

PR #236, opened 2022-03-14
Feat(NFT): Add a sort by Account | Collection button

darah3ng:
## Desc

- Added 2 buttons to sort the NFTs by account and collection.
- By default, the page will display the items sorted by account.
- If Collection is clicked, it will sort the items and display them by collection.

## Changes

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      <p>
        
      </p>
</td>
      <td>
    
  </td>
  </tr>
</table>
<p align='center'>

</p>

### Retrieving all NFTs
As far as I'm aware, the only way to get NFTs is by calling `useNFTsByAddress`, and you need to loop through all wallets to get all NFTs.
- This is tough when you're dealing with a feature like sorting where you need to have the total amount of items before you do the sorting. E.g. `total items > sort them > display sorted items`
- Also hooks will throw a hissy fit (breaking the rules of hooks) if you try to be a smart ass and wrap it in a utility function inside a functional component. (*I was that smart ass*)

Ideally, we want to be able to do something like `useNFTsByAddresses`. This way, you only need to feed the API all the wallets at once and retrieve all of the NFTs.

### Workaround-*ish*
- To get all NFTs on the NFTsPage component, I added a hook to accumulate NFTs as the page tries to render the NFTs from each account.
- Once it's done rendering (the accumulating should be done as well), we now have all the NFTs to do the sorting.
- The negative side is that this relies entirely on the initial rendering to complete, if it isn't, then we'll be looking at some unexpected behavior.
  - On Mr. Brideside, we're taking advantage of the initial render and collecting all NFTs, so we can use them to sort and display in a collection format without having to do another API call.

### Typescript

There are a few *cheeky* types `any` in this PR, which is not ideal since that defeats the purpose of using TS. I strongly recommend that they need to be typed out before this PR is merged into prod.

### Generic components

In NFTsPage, we're using `NFTGrid` and `NFTsBySort` which are very similar.
- We could potentially create a reusable component called `<NFTSection />`  that takes `account` or `collection` prop.
  - `account` prop for displaying `NFTGrid`
  - `collection` prop for displaying `NFTsBySort`
- This should shrink down the line numbers of the NFTsPage component and make it more readable.

## Bugs

1. `<NFTCard />` component is showing a blank image when trying to render video NFTs.

### DIFF
--- a/src/components/NFTsBySort.tsx
+++ b/src/components/NFTsBySort.tsx
@@ -1,0 +1,55 @@
+import { NftCard } from '@talisman-components/nft'
+import { device } from '@util/breakpoints'
+import styled from 'styled-components'
+
+const NFTsGridWrapper = styled.div`
+  display: grid;
+  gap: 2rem;
+  grid-template-columns: 1fr;
+  margin-bottom: 3rem;
+
+  @media ${device.md} {
+    grid-template-columns: repeat(2, 1fr);
+  }
+
+  @media ${device.lg} {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @media ${device.xl} {
+    grid-template-columns: repeat(4, 1fr);
+  }
+`
+
+export interface NFTsBySortProps {
+  sortedNfts: any[]
+}
+
+const NFTsBySort = ({ sortedNfts }: NFTsBySortProps) => {
+  if (!sortedNfts) {
+    return <h3>No sorted NFTs found!</h3>
+  }
+
+  const renderCollection = Object.keys(sortedNfts).map((collectionName: any) => {
+    return (
+      <div key={collectionName}>
+      <h2>{collectionName}</h2>
+
+      <NFTsGridWrapper>
+        {sortedNfts[collectionName].map((nft: any) => (
+            <NftCard key={nft.id} nft={nft} />
+          )
+        )}
+      </NFTsGridWrapper>
+    </div>
+    )
+  })
+
+  return (
+    <div>
+      {renderCollection}
+    </div>
+  )
+}
+
+export { NFTsGridWrapper, NFTsBySort }
--- a/src/components/NFTsBySort.tsx
+++ b/src/components/NFTsBySort.tsx
@@ -30,28 +30,23 @@ const NFTsBySort = ({ sortedNfts }: NFTsBySortProps) => {
     return <h3>No sorted NFTs found!</h3>
   }
 
-  console.log(typeof sortedNfts)
+  const renderCollection = Object.keys(sortedNfts)
+    .sort()
+    .map((collectionName: any) => {
+      return (
+        <div key={collectionName}>
+          <h2>{collectionName}</h2>
 
-  const renderCollection = Object.keys(sortedNfts).sort().map((collectionName: any) => {
-    return (
-      <div key={collectionName}>
-      <h2>{collectionName}</h2>
+          <NFTsGridWrapper>
+            {sortedNfts[collectionName].map((nft: any) => (
+              <NftCard key={nft.id} nft={nft} />
+            ))}
+          </NFTsGridWrapper>
+        </div>
+      )
+    })
 
-      <NFTsGridWrapper>
-        {sortedNfts[collectionName].map((nft: any) => (
-            <NftCard key={nft.id} nft={nft} />
-          )
-        )}
-      </NFTsGridWrapper>
-    </div>
-    )
-  })
-
-  return (
-    <div>
-      {renderCollection}
-    </div>
-  )
+  return <>{renderCollection}</>
 }
 
 export { NFTsGridWrapper, NFTsBySort }
--- a/src/components/NFTsBySort.tsx
+++ b/src/components/NFTsBySort.tsx
@@ -30,7 +30,9 @@ const NFTsBySort = ({ sortedNfts }: NFTsBySortProps) => {
     return <h3>No sorted NFTs found!</h3>
   }
 
-  const renderCollection = Object.keys(sortedNfts).map((collectionName: any) => {
+  console.log(typeof sortedNfts)
+
+  const renderCollection = Object.keys(sortedNfts).sort().map((collectionName: any) => {
     return (
       <div key={collectionName}>
       <h2>{collectionName}</h2>
--- a/src/components/SortButtons.tsx
+++ b/src/components/SortButtons.tsx
@@ -1,0 +1,35 @@
+import styled from 'styled-components'
+import Button from '@components/Button'
+
+const GroupSortsButton = styled.div`
+  display: flex;
+  padding: 0px 20px;
+  align-items: center;
+  margin-button: 40px;
+  justify-content: center;
+  gap: 10px;
+  color: black;
+`
+
+const SortButton = styled(Button)`
+  color: white;
+
+  ${({ isSelected, theme }) =>
+    !!isSelected &&
+    `
+      background: rgb(${theme?.primary});
+      color: rgb(${theme?.background});
+    `
+  }
+`
+
+const SortButtons = ({ setSortByAccount, setSortByCollection, sortBy }: any) => {
+  return (
+    <GroupSortsButton>
+      <SortButton isSelected={sortBy === 'account'} onClick={setSortByAccount}>Account</SortButton>
+      <SortButton isSelected={sortBy === 'collection'} onClick={setSortByCollection}>Collection</SortButton>
+    </GroupSortsButton>
+  )
+}
+
+export default SortButtons
--- a/src/components/SortButtons.tsx
+++ b/src/components/SortButtons.tsx
@@ -1,34 +1,37 @@
+import { Field } from '@components'
 import styled from 'styled-components'
-import Button from '@components/Button'
 
-const GroupSortsButton = styled.div`
+const SortButtonsWrapper = styled.div`
   display: flex;
   padding: 0px 20px;
-  align-items: center;
-  margin-button: 40px;
   justify-content: center;
   gap: 10px;
-  color: black;
+  color: var(--color-primary);
 `
 
-const SortButton = styled(Button)`
-  color: white;
+const sortOptions = [
+  {
+    key: 'account',
+    value: 'Account',
+  },
+  {
+    key: 'collection',
+    value: 'Collection',
+  },
+]
 
-  ${({ isSelected, theme }) =>
-    !!isSelected &&
-    `
-      background: rgb(${theme?.primary});
-      color: rgb(${theme?.background});
-    `
-  }
-`
-
-const SortButtons = ({ setSortByAccount, setSortByCollection, sortBy }: any) => {
+const SortButtons = ({ sortValue, setSortBy, setSortTrigger }: any) => {
   return (
-    <GroupSortsButton>
-      <SortButton isSelected={sortBy === 'account'} onClick={setSortByAccount}>Account</SortButton>
-      <SortButton isSelected={sortBy === 'collection'} onClick={setSortByCollection}>Collection</SortButton>
-    </GroupSortsButton>
+    <SortButtonsWrapper>
+      <Field.RadioGroup
+        value={sortValue}
+        onChange={(value: any) => {
+          setSortBy(value)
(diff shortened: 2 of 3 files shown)


juliuscarvajal:
Thanks for this submission @darah3ng. Really good initiative. The `@talisman-components/nft` is a private repo (for now) and was planning to create the `useNfts(addresses)` hook which gets all the Nfts for all the input addresses for the next iteration but did not have bandwidth.

Also the tricky bit of having a `useNfts` hook is that it would be tough to render errors retrieving NFTs for each address. Maybe the `useNfts` hook will return an array of errors for the UI to handle accordingly. Will need to think this more.

I'll have a look at this PR now.

src/routes/NFTsPage.tsx
```diff
@@ -81,37 +97,77 @@ const NFTGrid = styled(({ className = '', account }: AccountProps) => {
       line-height: 0;
     }
   }
+`
 
-  .nft-grid {
-    display: grid;
-    gap: 2rem;
-    grid-template-columns: 1fr;
+export enum SoryByValue {
```
juliuscarvajal, line 102: Typo.
I think it makes sense to just use this type def to avoid duplicate boilerplate:
```
export type SortByValue = 'account' | 'collection';
```

src/components/NFTsBySort.tsx
```diff
@@ -0,0 +1,57 @@
+import { NftCard } from '@talisman-components/nft'
+import { device } from '@util/breakpoints'
+import styled from 'styled-components'
+
+const NFTsGridWrapper = styled.div`
```
juliuscarvajal, line 5: Would be good to have this `NFTsGridWrapper` on a different file.

src/components/NFTsBySort.tsx
```diff
@@ -0,0 +1,57 @@
...
+export interface NFTsBySortProps {
+  sortedNfts: any[]
+}
+
+const NFTsBySort = ({ sortedNfts }: NFTsBySortProps) => {
+  if (!sortedNfts) {
+    return <h3>No sorted NFTs found!</h3>
+  }
+
+  console.log(typeof sortedNfts)
```
juliuscarvajal, line 33: Remove unnecessary `console.log`

src/components/NFTsBySort.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <div key={collectionName}>
+      <h2>{collectionName}</h2>
+
+      <NFTsGridWrapper>
+        {sortedNfts[collectionName].map((nft: any) => (
+            <NftCard key={nft.id} nft={nft} />
+          )
+        )}
+      </NFTsGridWrapper>
+    </div>
```
juliuscarvajal, line 46: Formatting seems off. Do you have prettier extension installed and setup?

src/components/NFTsBySort.tsx
```diff
@@ -0,0 +1,57 @@
...
+    )
+  })
+
+  return (
+    <div>
```
juliuscarvajal, line 51: Unnecessary wrapping div

src/components/SortButtons.tsx
```diff
@@ -0,0 +1,35 @@
+import styled from 'styled-components'
+import Button from '@components/Button'
+
+const GroupSortsButton = styled.div`
+  display: flex;
+  padding: 0px 20px;
+  align-items: center;
+  margin-button: 40px;
```
juliuscarvajal, line 8: You mean `margin-bottom`? Also use `rem` instead of `px`.
The root em (rem) is `10px` so meaning `40px === 4rem`.

src/components/SortButtons.tsx
```diff
@@ -0,0 +1,35 @@
...
+  justify-content: center;
+  gap: 10px;
+  color: black;
```
juliuscarvajal, line 11: Use css var `var(--color-text)` instead of `black`.

src/components/SortButtons.tsx
```diff
@@ -0,0 +1,35 @@
...
+const SortButtons = ({ setSortByAccount, setSortByCollection, sortBy }: any) => {
+  return (
+    <GroupSortsButton>
+      <SortButton isSelected={sortBy === 'account'} onClick={setSortByAccount}>Account</SortButton>
+      <SortButton isSelected={sortBy === 'collection'} onClick={setSortByCollection}>Collection</SortButton>
+    </GroupSortsButton>
+  )
+}
+
+export default SortButtons
```
juliuscarvajal, line 35: You can use the RadioGroup component for this.
Example taken from the `/crowdloans` page (`src/archetypes/CrowdloanV2/Indexx.tsx`)
```
          <Field.RadioGroup
            value={network}
            className="network-filter"
            onChange={(network: any) => {
              setNetwork(network)
              trackGoal('0AO7IT2G', 1) // crowdloan_filter
            }}
            options={networkOptions}
            small
            primary
          />
```

src/routes/NFTsPage.tsx
darah3ng: Ah nice catch, that's def a typo.

src/components/NFTsBySort.tsx
darah3ng: Hmm that's odd. Normally my prettier/eslint would throw a fat warning. 

Gonna need to check the config.

src/components/NFTsBySort.tsx
darah3ng: might be issue as mentioned [above](https://github.com/TalismanSociety/talisman-web/pull/236#discussion_r826799126)

src/components/SortButtons.tsx
darah3ng: Oopsie another typo, and good point on `rem`

src/components/NFTsBySort.tsx
darah3ng: I agree.

I placed it here because I wanted to get your input first. We can always refactor it later.

src/components/NFTsBySort.tsx
darah3ng: renderCollection in NFTsBySort returns an array of` [<div key={collectionName}></div>, <div key={collectionName}></div>, ....]`, therefore it needs to have a div wrapped on top so it can be a single root element.

On second thought, wrapping it with `<> </>` (fragment) seems to be a better approach here.

darah3ng:
@jonathanpdunne Thank you for your input. I just updated the code to align with the feedback.

Changes are:
- `console.log` removed
- CSS clean up
- Fix up some typo
- Update the `SortButton` component to use `Field.RadioGroup`

image

src/components/NFTsBySort.tsx
```diff
@@ -30,28 +30,23 @@ const NFTsBySort = ({ sortedNfts }: NFTsBySortProps) => {
...
-  })
-
-  return (
-    <div>
-      {renderCollection}
-    </div>
-  )
+  const renderCollection = Object.keys(sortedNfts)
+    .sort()
+    .map((collectionName: any) => {
```
juliuscarvajal, line 35: Good to have your prettier all sorted. Much easier to read code now :)
Try to avoid `any` if you can. Here, you can use `: string`.

src/components/SortButtons.tsx
```diff
@@ -1,34 +1,37 @@
...
-`
-
-const SortButtons = ({ setSortByAccount, setSortByCollection, sortBy }: any) => {
+const SortButtons = ({ sortValue, setSortBy, setSortTrigger }: any) => {
   return (
-    <GroupSortsButton>
-      <SortButton isSelected={sortBy === 'account'} onClick={setSortByAccount}>Account</SortButton>
-      <SortButton isSelected={sortBy === 'collection'} onClick={setSortByCollection}>Collection</SortButton>
-    </GroupSortsButton>
+    <SortButtonsWrapper>
```
juliuscarvajal, line 25: SortButtonsWrapper will introduce an extraneous div.
Try doing something like this:
```
const SortButtons = styled(({ className, sortValue, setSortBy, setSortTrigger }) => {
  return <Field.RadioGroup className={className} ...the-rest-of-the-props-here... />
})`
  ...the-styles...
`
```

juliuscarvajal: If above does not work then wrapper is ok

src/components/SortButtons.tsx
```diff
@@ -1,34 +1,37 @@
...
+      <Field.RadioGroup
+        value={sortValue}
+        onChange={(value: any) => {
```
juliuscarvajal, line 28: Use `string` instead of `any`.

src/routes/NFTsPage.tsx
```diff
@@ -99,34 +97,21 @@ const NFTGrid = styled(({ className = '', account, setTotalNfts }: AccountProps)
   }
 `
 
-export enum SoryByValue {
-  ACCOUNT = 'account',
-  COLLECTION = 'collection'
-}
+export type SoryByValue = 'account' | 'collection'
```
juliuscarvajal, line 100: Typo. Should be `SortByValue`.

src/routes/NFTsPage.tsx
```diff
@@ -99,34 +97,21 @@ const NFTGrid = styled(({ className = '', account, setTotalNfts }: AccountProps)
...
   const { accounts } = useExtensionAutoConnect()
   const [totalNfts, setTotalNfts] = useState<any[] | undefined>([])
   const [sortTrigger, setSortTrigger] = useState(false)
-  const [sortBy, setSortBy] = useState<SoryByValue>(SoryByValue.ACCOUNT)
+  const [sortBy, setSortBy] = useState<SoryByValue>('account')
 
   const updateTotalNfts = (nfts: any) => {
-    if(!nfts) return;
+    if (!nfts) return
     setTotalNfts((prevState: any) => {
```
juliuscarvajal, line 110: Try using `unknown` here instead of `any`. If TS still complains, then leave as-is.

src/components/NFTsBySort.tsx
```diff
@@ -0,0 +1,52 @@
...
+  @media ${device.xl} {
+    grid-template-columns: repeat(4, 1fr);
+  }
+`
+
+export interface NFTsBySortProps {
+  sortedNfts: any[]
+}
+
+const NFTsBySort = ({ sortedNfts }: NFTsBySortProps) => {
```
juliuscarvajal, line 28: This only sorts by collections but the NFTCard does not show the account address so would lose the context.
I think some UX input is needed here.

Ideally, accounts should still be shown but sorted by collections.

darah3ng:
@juliuscarvajal 

I did have concerns about [this](https://github.com/TalismanSociety/talisman-web/pull/236#discussion_r826896293) as I was building out the sort feature.
- One possible solution is to have a dropdown menu that follows the MVP design plus the sort buttons.
  - We then might need to modify the sort buttons to sort by NFTs or Collections
- This allows the users to filter to their individual accounts first and then, display them as either all NFTs or collections.
  -   **NFTs** - display all NFTs under their account (asc order)
  -  **Collections** - display all NFTs in a collection forrmat

Also, I intentionally used `any` in most places here since this was the first iteration and I wanted to confirm if everyone is happy with this approach before I proceed further. (should've specified this in the description 😅)

src/components/NFTsBySort.tsx
darah3ng: Turns out if you haven't closed your vscode for a while, it glitches out.

*I fixed it by restarting vscode* 🤯

juliuscarvajal:
Hey @darah3ng all good. I added a new hook `useNfts(addresses, options)` in `@talisman-components/nft` package.
This will probably mean that your workaround to gather all the NFTs might have to be ripped out. Apologies for this hook coming in too late.

The only concern for the `useNfts` hook is if the input addresses are too many and the collections returned are too much.

The second parameter `options` has `limit` and `limitPerAddress`.

Use one or the other where appropriate. If using both, `limit` will win over `limitPerAddress`.

Check out this PR with the sample usage of the new hook: https://github.com/TalismanSociety/talisman-web/pull/238

Please sync with the latest fork. Again, apologies for the inconvenience this has caused.

PR #273, opened 2022-05-02
Fix/nftversion2 issue fixes

JAhimaz:


jonathanpdunne (changes requested):
LGTM! 

Although I would remove that cheeky console.log in the NftModal.

jonathanpdunne (approved):

PR #302, opened 2022-09-29
build: yarn configs

tien:
Remove `checksumBehavior: update`, this is very important, as setting this to `update` enable a serious vector of attack where an attacker can add malicious code to a remote dependency while the app is being built & deployed. And later remove the malicious code from said dependency without anyone knowing it

Also set Yarn version to latest stable release & clean up other minor configs.


alecdwm (approved):
Looking good! 🔥

PR #303, opened 2022-09-29
chore: organize dependencies used for development under `devDependencies`

tien:
Although technically, this doesn't have any actual impact for bundled application. It would still help us to easily identify the key roles of different dependencies.


alecdwm:
Hell yeah!

Do you think `@open-wc/webpack-import-meta-loader` should also be included in devDependencies?
Although, I'm not even sure if we're using it to be honest

tien:
> @open-wc/webpack-import-meta-loader

Oops missed that one, have updated accordingly, I'll try to identify all dependencies' usage in a separate PR 💪

alecdwm (approved):
Looking good!

PR #304, opened 2022-09-29
build: set project targeted node version

tien:
This is especially useful for Yarn 2. With random bugs arising from using unsupported node version.


alecdwm (approved):
🎉

PR #305, opened 2022-09-29
chore: clean up prettier workflow

tien:
# Changes:

## [chore: use pretty-quick for running Prettier on staged files](https://github.com/TalismanSociety/talisman-web/commit/2da4cf4700dc6dad51bbc6f151b08d732691cef3)

`pretty-quick` automatically detect & format file types supported by Prettier,
so we don't have to list every possible file extensions when using `lint-staged`
json, yaml, xml, etc should work right out of the box

## [ci: work flow for enforcing correctly formatted files using Prettier](https://github.com/TalismanSociety/talisman-web/commit/29f09959049da5390c8bf2b4b10f7d46a1568b32)

Delete old work flow where formatting also happen on CI server.
Developer should make a concious decision of what they have committed
& CI should only enforce code quality check & not modifying the actual PR


.github/workflows/validation.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 'lts/*'
+          cache: yarn
+
+      - run: yarn install --immutable
+      - run: yarn pretty-quick --check
```
alecdwm, line 22: Awesome 🔥

alecdwm (approved):
Love it

PR #310, opened 2022-10-03
feat: Adding ORML balances to the Web-app

JAhimaz:
This feature branch adds the ORML balances libs into the the web app.


src/archetypes/Wallet/Assets.tsx
```diff
@@ -233,3 +154,10 @@ const ExtensionUnavailable = styled(props => {
     font-size: 1.6rem;
   }
 `
+
+function useAddressesByToken(addresses: string[] | null | undefined, tokenIds: Token['id'][]) {
+  return useMemo(() => {
+    if (addresses === undefined || addresses === null) return {}
```
tien, line 160: I noticed we have `lodash` as dependency, so can use the utility function `isNil` to shorten this.

src/archetypes/Wallet/NewBalanceExample.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const addressesByToken = useAddressesByToken(addresses, tokenIds)
+  const balances = useBalances(balanceModules, chaindata, addressesByToken)
+
+  return (
+    <>
+      <h2>Balances Demo</h2>
+
+      {/* Display balances per balance (so, per token per account) */}
+      {balances?.sorted.map(balance =>
+        balance.total.planck === BigInt('0') ? null : (
```
tien, line 37: - `BigInt` take a number primitive, so we can do `Bigint(0)` instead.
- Since `balance.total.planck === BigInt('0')` is an explicit conditional (evaluate to boolean), we can do `balance.total.planck === BigInt('0') && (<SomeComponent />)` without having to use ternary operator.

src/archetypes/Wallet/NewBalanceExample.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <div key={balance.id} style={{ display: 'flex', alignItems: 'center', gap: '1rem' }}>
+            <img alt="token logo" src={balance.token?.logo} style={{ height: '2rem', borderRadius: '9999999rem' }} />
```
tien, line 39: I don't have enough context here, but we can also achieve a round border by using `borderRadius: '100%'`

src/archetypes/Wallet/NewBalanceExample.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+            <span>{balance.status}</span>
+
+            <span>{balance.chain?.name || balance.evmNetwork?.name}</span>
```
tien, line 43: Same as above, better to use `??` here.

src/archetypes/Wallet/NewBalanceExample.tsx
```diff
@@ -0,0 +1,113 @@
...
+              {formatDecimals(balance.transferable.tokens)} {balance.token?.symbol}
+            </span>
+            <span style={{ opacity: '0.6', fontSize: '0.8em' }}>
+              {typeof balance.transferable.fiat('usd') === 'number'
+                ? new Intl.NumberFormat(undefined, {
+                    style: 'currency',
+                    currency: 'usd',
+                    currencyDisplay: 'narrowSymbol',
+                  }).format(balance.transferable.fiat('usd') || 0)
+                : ' -'}
```
tien, line 54: Same as above, this could also benefit from using `toLocaleString` & nullish coalescing.

src/archetypes/Wallet/NewBalanceExample.tsx
```diff
@@ -0,0 +1,113 @@
...
+ *       [tokenIdTwo]: [addressOne, addressTwo, etc]
+ *       [etc]:        [addressOne, addressTwo, etc]
+ *     }
+ */
+function useAddressesByToken(addresses: string[] | null | undefined, tokenIds: Token['id'][]) {
+  return useMemo(() => {
+    if (addresses === undefined || addresses === null) return {}
+    return Object.fromEntries(tokenIds.map(tokenId => [tokenId, addresses]))
+  }, [addresses, tokenIds])
+}
```
tien, line 113: I just realized that this function is duplicated of couple times. Is it possible to have this extracted to a shared hook?

src/archetypes/Wallet/Total.tsx
```diff
@@ -2,53 +2,56 @@ import { Pendor, Pill } from '@components'
...
+  const addressesByToken = useAddressesByToken(addresses, tokenIds)
+  const balances = useBalances(balanceModules, chaindata, addressesByToken)
+
+  const assetsValue = typeof balances?.sum.fiat('usd').transferable === 'number'
+  ? new Intl.NumberFormat(undefined, {
+    style: 'currency',
+    currency: 'usd',
+    currencyDisplay: 'narrowSymbol',
+  }).format(balances?.sum.fiat('usd').transferable || 0)
+  : <Loader />
```
tien, line 47: Same as above, this could also benefit from using `toLocaleString` & nullish coalescing.

tien (changes requested):
I don't have context over the actual requirements, but looks good overall 💪
Only got a couple of comments

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,77 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
-const AssetBalance = styled(({ id, balances, addresses }) => {
-  const { status, accounts } = useExtensionAutoConnect()
-  const { parachainDetails } = useParachainDetailsById(id)
 
-  const isMoonriver = ['Moonriver', 'Moonbeam'].includes(parachainDetails?.name as string)
-  const hasNoEthereumAddress = useMemo(
-    () => status === 'OK' && accounts.every(account => account.type !== 'ethereum'),
-    [status, accounts]
-  )
+  if(tokenAmount === '0') return null
```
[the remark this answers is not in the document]
JAhimaz: Ah this was a test for when we were trying to figure out the BigInt issue. it has been reverted.

src/archetypes/Wallet/NewBalanceExample.tsx
JAhimaz: @alecdwm possibly include in the library as well?

src/archetypes/Wallet/NewBalanceExample.tsx
alecdwm: I'm not sure if it's relevant to this specific case, but `100%` behaves differently to `9999999rem` when the element can't fit. For example:

Screen Shot 2022-10-05 at 01 09 51

[more deets](https://stackoverflow.com/a/29966500/3926156)

tien: Oh damn, actually didn't think of this case before. Yeah, ignore if this applied here

package.json
```diff
@@ -19,11 +19,19 @@
     "@polkadot/util-crypto": "^7.4.1",
     "@ramp-network/ramp-instant-sdk": "^2.5.0",
     "@talismn/api-react-hooks": "~0",
```
alecdwm, line 21: Let's ditch this as part of this PR! ✨

alecdwm: (`@talismn/api-react-hooks` and `@talismn/chaindata-js`)

JAhimaz: Might want to clean it up a bit as some stuff still utilises `@talismn/api-react-hooks`. On the otherhand, `@talismn/chaindata-js` is not used except in crowdloans but the merge should deprecate that.

src/archetypes/Wallet/Assets.tsx
JAhimaz:
```
function useAddressesByToken(addresses: string[] | null | undefined, tokenIds: Token['id'][]) {
  return useMemo(() => {
    if (isNil(addresses)) return {}
    return Object.fromEntries(tokenIds.map(tokenId => [tokenId, addresses]))
  }, [addresses, tokenIds])
}
```

Done :)

JAhimaz (commented):
It's pretty outdated now, but I've updated it with most of the changes you had commented on @tien

src/archetypes/Account/Button.tsx
```diff
@@ -92,19 +94,36 @@ const Dropdown = styled(
...
           {totalBalanceByAddress === '0' && showBuy && <BuyItem nativeToken={nativeToken} onClick={closeParent} />}
           {(allAccounts ? [{ name: t('All Accounts') }, ...accounts] : accounts).map(
             ({ address, name, type, genesisHash }, index) => {
-              const totalBalance = totalBalanceByAddressFunc(address)
+              // const { assetsValue } = _useBalances()
+              const { assetsValue, balances } = _useBalances()
+
+              // Do the filtering
+
+              let fiatBalance = null
```
tien, line 114: Rather than using `let` & mutation here, we could use a ternary operator on `address` instead.

src/archetypes/Account/Button.tsx
```diff
@@ -155,16 +174,13 @@ const Dropdown = styled(
                     {address ? (
                       // show usd when no chainId specified
                       parachainId === undefined ? (
-                        <Pendor prefix={!totalUsdByAddress[encodeAnyAddress(address, 42)] && '-'}>
-                          {totalUsdByAddress[encodeAnyAddress(address, 42)] &&
-                            formatCurrency(totalUsdByAddress[encodeAnyAddress(address, 42)])}
-                        </Pendor>
+                        <Pendor prefix={!fiatBalance && '-'}>{fiatBalance}</Pendor>
```
tien, line 177: If `prefix` evaluate to `false`, will we end up with a string "false" here?

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
-  const { t } = useTranslation()
-  const { parachainDetails } = useParachainDetailsById(id)
-  const asset = parachainDetails?.name
+type AssetBalanceProps = {
+  token: Token
+  balances: Balances | undefined
+  address: string | undefined
+}
+const AssetBalance = styled(({ token, balances, address }: AssetBalanceProps) => {
+  let tokenBalances = null
```
tien, line 41: Same here, avoid using `let` and mutation if possible.

src/archetypes/Account/Button.tsx
```diff
@@ -92,19 +94,36 @@ const Dropdown = styled(
...
+              if (!address) fiatBalance = assetsValue
+              else
+                fiatBalance =
+                  typeof balances?.find({ address: address })?.sum?.fiat('usd').transferable === 'number'
+                    ? new Intl.NumberFormat(undefined, {
+                        style: 'currency',
+                        currency: 'usd',
+                        currencyDisplay: 'narrowSymbol',
+                      }).format(balances?.find({ address: address })?.sum?.fiat('usd').transferable || 0)
+                    : '-'
```
tien, line 125: - When check the type definition for `transferable` it's a `number` only, is the type definition for that incorrect hence we have to check for type `number` here?
- `transferable || 0` could be `transferable ?? 0` instead, for explicit null check.

I believe this would be much cleaner

```tsx
(balances?.find({ tokenId: token.id }).sum.fiat('usd').transferable ?? 0)
  .toLocaleString(undefined, {
    style: 'currency',
    currency: 'USD',
    currencyDisplay: 'narrowSymbol',
  }) ?? ' -';
```

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
-  const shouldShowInstructions = isMoonriver && hasNoEthereumAddress
-  const isNetworkId = id === '0' || id === '2'
+  const fiatAmount =
+    typeof tokenBalances.sum.fiat('usd').transferable === 'number'
+      ? new Intl.NumberFormat(undefined, {
+          style: 'currency',
+          currency: 'usd',
+          currencyDisplay: 'narrowSymbol',
+        }).format(tokenBalances.sum.fiat('usd').transferable || 0)
+      : '-'
```
tien, line 59: Same as above, see my comment on line: `src/archetypes/Account/Button.tsx:119`

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
-    () =>
-      Object.entries(totalAssetsUsdByAddress || {})
-        .filter(([address]) => accountAddresses && accountAddresses.includes(address))
-        .map(([, assetsUsd]) => assetsUsd)
-        .reduce(addBigNumbers, undefined),
-    [totalAssetsUsdByAddress, accountAddresses]
-  )
+
+  const { balances, tokenIds, tokens, assetsValue } = useBalances()
+  const address = useActiveAccount().address
```
tien, line 83: Can do this instead
```ts
const { address } = useActiveAccount();
```

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
+  const chaindata = useChaindata()
+
+  const chains = useChains(chaindata)
+  const evmNetworks = useEvmNetworks(chaindata)
+
+  let fiatTotal: string | null = ''
```
tien, line 89: Same as above, try to avoid using `let` & mutation.

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
+  if (!address) fiatTotal = assetsValue
+  else
+    fiatTotal =
+      typeof balances?.find({ address: address })?.sum?.fiat('usd').transferable === 'number'
+        ? new Intl.NumberFormat(undefined, {
+            style: 'currency',
+            currency: 'usd',
+            currencyDisplay: 'narrowSymbol',
+          }).format(balances?.find({ address: address })?.sum?.fiat('usd').transferable || 0)
+        : '-'
```
tien, line 100: Same as above, see my comment on line: `src/archetypes/Account/Button.tsx:119`.

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
+
+  const value = balances?.find({ address: address })?.sum?.fiat('usd').transferable
 
   return (
     <section className={`wallet-assets ${className}`}>
-      <Panel title={t('Assets')} subtitle={assetsUsd && formatCurrency(assetsUsd)}>
+      <Panel title={t('Assets')} subtitle={!balances ? <Loader /> : fiatTotal}>
```
tien, line 106: In this case `!balances` could give the wrong behaviour as if user doesn't have any balance & loading has finished: `![]` will evaluate to false.

@JAhimaz @alecdwm but in all cases, it's always better to do explicit check anyway `someValue !== undefined` rather than `!someValue`

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
+                const aCmp = aChain?.name?.toLowerCase() || a.chain?.id || a.evmNetwork?.id
+                const bCmp = bChain?.name?.toLowerCase() || b.chain?.id || b.evmNetwork?.id
+
+                if (aCmp === undefined && bCmp === undefined) return 0
+                if (aCmp === undefined) return 1
+                if (bCmp === undefined) return -1
+
+                return aCmp.localeCompare(bCmp)
+              })
+              .map(token => <AssetBalance key={token.id} token={token} balances={balances} address={address} />)
```
tien, line 142: For complex operation like this, better to extract this out into a `useMemo`, so that this doesn't get run on every render.

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
+          ) : (
+            <AssetsLoading />
+          )}
+          {value === 0 && !!address ? (
```
tien, line 146: `!!address` lots of implicit checks like this throughout the PR. @JAhimaz can you find and replace them with explicit check if possible?

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
+            <>
+              <PanelSection>
+                <AssetItem
+                  token={tokens['kusama-substrate-native-ksm']}
+                  tokenAmount={'0'}
+                  fiatAmount={new Intl.NumberFormat(undefined, {
+                    style: 'currency',
+                    currency: 'usd',
+                    currencyDisplay: 'narrowSymbol',
+                  }).format(0)}
```
tien, line 156: Better to use `toLocaleString` here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

src/archetypes/Wallet/Assets.tsx
```diff
@@ -96,99 +32,147 @@ const AssetItem = styled(({ id, balances, addresses, className }) => {
...
+              </PanelSection>
+              <PanelSection>
+                <AssetItem
+                  token={tokens['polkadot-substrate-native-dot']}
+                  tokenAmount={'0'}
+                  fiatAmount={new Intl.NumberFormat(undefined, {
+                    style: 'currency',
+                    currency: 'usd',
+                    currencyDisplay: 'narrowSymbol',
+                  }).format(0)}
```
tien, line 169: Same as above, unless for specific case, or when we want to share the `NumberFormat` instance better to use [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString)

src/archetypes/Wallet/Assets.tsx
tien: Also with explicit comparison, you don't have to use ternary operator any more, just `&&` would be sufficient.

src/archetypes/Wallet/Assets.tsx
```diff
@@ -201,10 +185,65 @@ const Assets = styled(({ className }) => {
...
+      max-height: 400px;
+      overflow-y: auto;
+    }
+  }
 `
 
 export default Assets
 
+const AssetsLoading = styled(({ className }) => {
+  const placeholderAssets = ['🍜', '🍜', '🍜', '🍜', '🍜']
```
tien, line 200: This is fine, but just food for thought we can also do this :)
```ts
const placeholderAssets = Array(5).fill('🍜');
```

src/archetypes/Wallet/Assets.tsx
```diff
@@ -201,10 +185,65 @@ const Assets = styled(({ className }) => {
...
+
+  @keyframes shimmer {
+    0% {
+      background-position: -1000px 0;
+    }
+    100% {
+      background-position: 1000px 0;
+    }
+  }
+`
```
tien, line 245: If you rebase from `prod`. We have `emotion` as opposed to `styled-components`. So you can start declaring CSS as in-line object. Would be good if you can try it out and get a hang of it.

reference: https://emotion.sh/docs/object-styles

src/archetypes/Wallet/Assets.tsx
```diff
@@ -233,3 +272,22 @@ const ExtensionUnavailable = styled(props => {
...
+ *     }
+ */
+
+function getNetworkType({ chain, evmNetwork }: Balance): string | null {
+  if (evmNetwork) return evmNetwork.isTestnet ? 'EVM Testnet' : 'EVM Blockchain'
+  if (chain) {
+    if (chain.isTestnet) return 'Testnet'
+    return chain.paraId ? 'Parachain' : (chain.parathreads || []).length > 0 ? 'Relay Chain' : 'Blockchain'
+  }
+  return null
```
tien, line 292: It's always better to return unhappy path early:
- Always cover for error, unhappy path first
- Allow us to reduce nesting level

like so: `if (chain === undefined) return undefined`

Also, `chain.parathreads || []` can be `chain.parathreads ?? []`

src/archetypes/Wallet/Total.tsx
```diff
@@ -1,54 +1,34 @@
...
-      Object.entries(totalUsdByAddress || {})
-        .filter(([address]) => accountAddresses && accountAddresses.includes(address))
-        .map(([, usd]) => usd)
-        .reduce(addBigNumbers, undefined),
-    [totalUsdByAddress, accountAddresses]
-  )
+  const { balances, assetsValue } = useBalances()
+  const address = useActiveAccount().address
+
+  let fiatTotal: string | null = null
```
tien, line 13: Same as above, try to avoid using `let` and mutation.

src/archetypes/Wallet/Total.tsx
```diff
@@ -1,54 +1,34 @@
...
+  if (!address) fiatTotal = assetsValue
+  else
+    fiatTotal =
+      typeof balances?.find({ address: address })?.sum?.fiat('usd').transferable === 'number'
+        ? new Intl.NumberFormat(undefined, {
+            style: 'currency',
+            currency: 'usd',
+            currencyDisplay: 'narrowSymbol',
+          }).format(balances?.find({ address: address })?.sum?.fiat('usd').transferable || 0)
+        : '-'
```
tien, line 24: Same as above, see my comment on line: `src/archetypes/Account/Button.tsx:119`.

src/archetypes/Wallet/Total.tsx
```diff
@@ -1,54 +1,34 @@
...
-            {totalUsdChange && (
-              <Pill primary small>
-                {formatCurrency(totalUsdChange)}
-              </Pill>
-            )}
-          </>
-        </Pendor>
-      )}
+      <div className="amount">
+        <span>{!!fiatTotal ? fiatTotal : <StyledLoader />}</span>
```
tien, line 30: can do `fiatTotal ?? <StyledLoader />`

src/components/Panel.tsx
```diff
@@ -34,17 +34,35 @@ export const Section = styled(({ title, children, className, comingSoon, ...rest
...
+  //     }
+  //   }
+  // };
+
+  return (
+    <div className={`panel ${className}`} {...rest}>
+      {!!title && (
+        <h1>
+          {title}
+          {!!subtitle && <span>{subtitle}</span>}
```
tien, line 59: Better to do explicit condition here rather than using double bang `!!`

src/layout/StateBanner/index.tsx
```diff
@@ -1,14 +1,19 @@
...
 import { ExploreCrowdloansBanner } from './ExploreCrowdloansBanner'
 import { NoWalletBanner } from './NoWalletBanner'
 
 export const StateBanner = () => {
-  const hasEmptyBags = usePortfolioHasEmptyBags()
-  const { status } = useActiveAccount()
+  const { status, address } = useActiveAccount()
+  const { assetsValue, balances } = useBalances()
+
+  let hasEmptyBags = false
```
tien, line 12: Same as above, try to avoid using `let` and mutation.

src/libs/portfolio/index.tsx
```diff
@@ -224,7 +225,8 @@ export const Provider: FC<ProviderProps> = ({ children }) => {
     })
 
     const isLoading = Object.keys(loadingList).length !== 0
-    const hasEmptyBags = !isLoading && totalUsd === '0'
+    // const hasEmptyBags = !isLoading && totalUsd === '0'
+    const hasEmptyBags = false
```
tien, line 229: Can we double-check if this is the desired constant?

src/libs/talisman/balances.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const Context = createContext<ContextProps>({
+  balances: undefined,
+  assetsValue: '',
+  tokenIds: [],
+  tokens: [],
+  chaindata: null,
+})
+
+function useContext() {
```
tien, line 40: Can we rename this to `useBalanceContext` and leave `React.useContext` as is, to avoid confusion.

src/libs/talisman/balances.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const balances = _useBalances(balanceModules, chaindata, addressesByToken)
+
+  const assetsValue =
+    typeof balances?.sum.fiat('usd').transferable === 'number'
+      ? new Intl.NumberFormat(undefined, {
+          style: 'currency',
+          currency: 'usd',
+          currencyDisplay: 'narrowSymbol',
+        }).format(balances?.sum.fiat('usd').transferable || 0)
+      : null
```
tien, line 78: Same as above, see my comment on line: `src/archetypes/Account/Button.tsx:119`

tien (changes requested):
Got a couple more comments. Feel free to reach out if you have any questions :D

src/archetypes/Account/Button.tsx
JAhimaz: @alecdwm

src/archetypes/Wallet/Assets.tsx
JAhimaz: Will do! Will start using emotion from now on.

tien (approved):
🚢 it 🔥

JAhimaz:
Away we go 🚀

PR #316, opened 2022-10-05
chore: fix formatting issues

tien:
- Formatted some files that we missed
- Fix CI step for checking Prettier: GitHub Action checkout in detached mode, so we can't use `pretty-quick`, since that will try to find the best common ancestor(s) between two commits to run the check, which doesn't exist when the branch is detached


alecdwm (approved):
🤙

PR #317, opened 2022-10-06
chore: upgrade React to v18

tien:
A very important milestone, especially with the performance boost introduced by [Concurrent React](https://reactjs.org/blog/2022/03/29/react-v18.html#what-is-concurrent-react).

https://reactjs.org/blog/2022/03/29/react-v18.html

TODO: upgrade `react-scripts`, currently version 5 with its dependency on webpack 5 is causing some issues.


JAhimaz (approved):
lgtm

PR #321, opened 2022-10-10
feat: components for nomination pools

tien:
- Set up storybook
- Atomics design system components structure
- Create some basics components needed for nomination pools


JAhimaz (approved):
lgtm

PR #325, opened 2022-10-17
feat: nomination pools

tien:


src/archetypes/Transaction/Item.tsx
```diff
@@ -87,7 +87,7 @@ export const Item = styled(({ className, transaction, addresses, selectedAccount
                 {truncateAddress(youAddress ? encodeAnyAddress(youAddress, ss58Format) : youAddress, 4)}
               </ClickToCopy>
             }
-            graphic={<Avatar value={youAddress} />}
+            graphic={<Identicon value={youAddress} size="3.2rem" />}
```
alecdwm, line 90: 🤩

src/domains/chains/recoils.ts
```diff
@@ -0,0 +1,116 @@
...
+  subscanUrl: string | null
+}
+
+const SUPPORTED_CHAIN_IDS = ['polkadot', 'kusama', 'westend-testnet']
+
+export const chainsState = selector({
+  key: 'Chains',
+  get: async () => {
+    const response = await request<{ chains: Chain[] }>(
+      'https://app.gc.subsquid.io/beta/chaindata/v3/graphql',
```
alecdwm, line 28: Would be good to stick this string in a config file or maybe an ENV var!
It's likely that `v3` will need to change in the future, and it's also pretty likely that the baseurl will be different in the future.

src/domains/chains/recoils.ts
```diff
@@ -0,0 +1,116 @@
...
+  get:
+    (fiat: string = 'usd') =>
+    async ({ get }) => {
+      const chain = get(currentChainState)
+
+      if (chain.isTestnet) return 1
+
+      try {
+        const result = await fetch(
+          `https://api.coingecko.com/api/v3/simple/price?ids=${chain.nativeToken.data.coingeckoId}&vs_currencies=${fiat}`
```
alecdwm, line 80: Would be good to batch the queries to this endpoint (by passing multiple coingecko ids like in [here](https://github.com/TalismanSociety/talisman/blob/df71983e47ac720e2e46471890d101f59e1e7fce/packages/token-rates/src/TokenRates.ts#L50-L53) when/if at all possible).

I think we plan to get a coingecko subscription, but even with one there's still not a very large limit to the number of requests we can make.  
It would be good to save them for use cases like historical data lookup (which is impossible to batch) rather than use it on batchable queries.

alecdwm (approved):
Cool stuff!

Left some comments, but nothing crazy

tien:
@alecdwm thanks for the feedback regarding coingecko n subquid graphql, will have those fixed during the cooldown 💪, gonna get this out to prod first

PR #354, opened 2022-11-02
build: migrate `craco-alias` & other clean-ups

tien:


alecdwm (approved):
Looks good!

One note: I really liked how we used to split this file up into individual plugins.  
It made it easy to see which properties were supposed to go together, so one could add or remove a piece of functionality without risk of breaking the other ones.

tien:
> Looks good!
> 
> 
> 
> One note: I really liked how we used to split this file up into individual plugins.  
> 
> It made it easy to see which properties were supposed to go together, so one could add or remove a piece of functionality without risk of breaking the other ones.

True, ideally we want to do all away with these configs though. Once he get react script to v5 a lot of these shouldn't be needed anymore I think.

alecdwm:
Hah, yup that will be even better!

PR #356, opened 2022-11-03
chore: remove legacy util lib

tien:


src/archetypes/Wallet/Crowdloans.tsx
```diff
@@ -34,15 +34,13 @@ const CrowdloanItem = styled(({ id, className }) => {
   const { contributions } = useCrowdloanContributions({ accounts, crowdloans: id ? [id] : undefined })
   const totalContributions = getTotalContributionForCrowdloan(id, contributions)
 
-  const relayTokenSymbol = useFuncMemo(token => token || 'Planck', relayNativeToken)
-  const contributedTokens = useFuncMemo(planckToTokens, totalContributions || undefined, relayTokenDecimals)
-  const contributedUsd = multiplyBigNumbers(contributedTokens, relayTokenPrice)
+  const relayTokenSymbol = relayNativeToken ?? 'Planck'
+  const contributedTokens = useMemo(() => planckToTokens(totalContributions || undefined, relayTokenDecimals), [])
```
alecdwm, line 38: I think the equivalent code to what was there before would be:

```typescript
  const contributedTokens = useMemo(
    () => planckToTokens(totalContributions || undefined, relayTokenDecimals),
    [planckToTokens, totalContributions, relayTokenDecimals]
  )
```

Instead of with an empty dependency array

Same for the other places where `useFuncMemo` used to be used

alecdwm (changes requested):
Looks good! But I think there might be one thing to do with `useFuncMemo`

src/archetypes/Wallet/Crowdloans.tsx
tien: oh yeah, eslint wasn't catching those for some reason, lemme fix those real quick

tien: Oh wait it does, I'm just blind :)

alecdwm (approved):
🎉

PR #359, opened 2022-11-03
chore: posthog improvements

tien:


JAhimaz (approved):
LGTM. Will work on the required click throughs as per request of Birbo

PR #360, opened 2022-11-03
feat: improved nft modal

JAhimaz:


src/archetypes/NFT/Card/Card.tsx
```diff
@@ -19,6 +19,8 @@ function Card({ className, nft }: CardProps) {
     <>
       {open && (
         <AlertDialog
+          title={nft.name ?? nft.id}
+          subtitle={nft?.collection?.name ?? nft?.provider}
           open={true}
           content={<Modal id={nft.id} />}
           onRequestDismiss={() => setOpen(false)}
```
tien, line 26: For modal component, it's better to pass in the open prop. Rather than mount and unmount directly. Reason being: component like this could have exit animation. If you do this only, enter animation will be play (currently AlertDialog doesn't have exit animation so it actually no different). Same thing would go for stuff like dropdown, tooltip, etc

src/archetypes/NFT/Modal/LoadingState.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <span className="title" />
+      {/* Array of 4 span with className description */}
+      {[...Array(5)].map((_, i) => (
+        <span key={i} className="description" />
+      ))}
+    </LoadingTiles>
+  )
+}
+
+const LoadingTiles = styled.div`
```
tien, line 15: - Try using [in-line object styling](https://emotion.sh/docs/object-styles) for this instead
- Just FYI, key frames can also be achieved using emotion https://emotion.sh/docs/keyframes. But for this loading animation, take a look at `Skeleton` component under `@components/atoms`. Currently, this isn't documented using Storybook yet, and `Skeleton.Foreground`, doesn't have a shimmer animation. Would be a great exercise for you to do 💪

tien (changes requested):
Looks good, got only 2 comments

tien (approved):
👌

PR #376, opened 2022-11-07
feat: re-enable crowdloan

### LINKED ISSUE #138: Rehydrate portfolio crowdloan contribution on successful contribution
We should requery for crowdloan contributions after successfully contributing - may require a delay of 5-10 seconds?

Another option is to manually inject the crowdloan contribution into the store, so it's there, then manually run the query after 5-10 sec. 

Currently it's rehydrating at a set interval, we could probably remove that and just refresh after contribution.

### LINKED ISSUE #290: Crowdloans no longer have an ended status.
The status-based filtering on crowdloans needs to be updated. This may require looking at the endedAt field to determine when one has ended.

tien:
Fix #290 
Fix #138


jonathanpdunne (changes requested):
image

Would address this on the contribute modal.

SwamiTLSM (approved):

PR #378, opened 2022-11-08
Fix/type errors

jonathanpdunne:


tien (changes requested):
Thanks for all the 🐜🔨

Just 2 things:

- The added `className` props should also be optional `{ className?: string }`
- Parent class name should not be pass onto child elements, since parent style could leak into all of them:
```tsx
const Child = (props: { className?: string }) => <div className={props.className} />;

const Parent = (props: { className?: string }) => (
  <div className={props.className}>
    <Child className={props.className} />
    <Child className={props.className} />
    <Child className={props.className} />
  </div>
);

/**
 * Expect a box with red border
 * but we end up with 3 red border boxes inside 1 red border one
 */
export const Main = () => (
  <Parent
    css={{
      border: '1px solid red',
    }}
  />
);
```

PR #380, opened 2022-11-08
refactor: support type inference for storage call containing `GenericStorageEntryFunction`

tien:


tien:
> Generally prefer more verbose naming like `TSection` & `TModule` instead of `S` & `M`. But each to their own: ). Otherwise this is awesome!

It's mainly to differentiate between the generic type `TSection` & the inferred type ` infer S`. But yeah good point, I'll change it to `Section` without the `T` instead

jonathanpdunne (approved):
LGTM

PR #384, opened 2022-11-09
fix: resolved issue with statemine, acala and type errors

JAhimaz:


apps/web/src/libs/@talisman-nft/providers/Acala/AcalaProvider.ts
```diff
@@ -47,22 +40,20 @@ export class AcalaProvider extends NFTInterface {
...
-    const tokenDetails = await this.webSocket.query.ormlNft.tokens(collectionId, nftTokenId)
-    tokenDetails.toHuman() as unknown as Token
-    const collectionDetails = (await this.webSocket.query.ormlNft.classes(collectionId)).toHuman() as Record<
-      string,
-      any
-    >
+    const collectionIdFixed: string = collectionId.replaceAll(',', '')
+    const nftTokenIdFixed: string = nftTokenId.replaceAll(',', '')
+
+    const collectionDetails = (await this.webSocket.query.ormlNFT.classes(collectionIdFixed)).unwrap()
```
tien, line 46: `unwrap` will throw an error if the value is empty. Can do this instead:

```ts
const maybeCollectionDetails = await this.webSocket.query.ormlNFT.classes(collectionIdFixed);

if (collectionDetails.isNone) return null;

const collectionDetails = maybeCollectionDetails.unwrap();
```

tien (changes requested):
Just the single comment, everything else look allg 👌

apps/web/src/libs/@talisman-nft/providers/Evm/EVMProvider.ts
```diff
@@ -113,10 +93,36 @@ export class EVMProvider extends NFTInterface {
...
+      const chainId = await provider
+        .getNetwork()
+        .then(network => network.chainId)
+        .catch(e => undefined)
+      if (chainId !== this.chainId || chainId === undefined) {
+        continue
+      }
+      this.web3 = provider
+      break
+    }
```
tien, line 123: Sequential asynchronous blocking calls like this are very slow. Try using Promise utilities like `Promise.all` instead.

tien (changes requested):
Just these 2 issues

tien (approved):
Lgtm

PR #389, opened 2022-11-11
test: basic sanity checks

tien:
Right now, this test only performs basic sanity checks by iterating over all pages and check for empty state

TODO: Integrate mock account/extension connection


jonathanpdunne (changes requested):
Couple of conflicts to resolve, but otherwise looks good!

jonathanpdunne (approved):

PR #394, opened 2022-11-17
tx history improvements

alecdwm:


apps/web/src/archetypes/Transaction/ItemNoDetails.tsx
```diff
@@ -29,6 +29,11 @@ export const ItemNoDetails = () => {
       >
         {t(`We don't recognise this\ntransaction type yet...`)}
       </div>
+      <div
+        css={css`
+          width: 36px;
+        `}
```
tien, line 35: Would be better to use object notation. Also rem instead of px.

tien (dismissed):
Just got the single comment

SwamiTLSM (approved):

PR #406, opened 2022-11-24
fix: pool status indicator

tien:
# Fixed

- Tool tip was hidden on staking input
- On portfolio page, initial status was `not_earning_rewards`


jonathanpdunne (approved):
LGTM

PR #407, opened 2022-11-24
fix: fixed issues with nfts attributes trying to access objects within objects

JAhimaz:
Attributes on some NFTs have objects as their values, this temporary fix checks if the value is a string. If it isn't return `unknown`.

No NFT standards in the dotworld 😢

### DIFF
--- a/apps/web/src/archetypes/NFT/Modal/NftDialog.tsx
+++ b/apps/web/src/archetypes/NFT/Modal/NftDialog.tsx
@@ -39,6 +39,8 @@ const backdropKeyframes = keyframes`
 const NftDialog = (props: NftDialogProps) => {
   const { nft, loading } = useNftById(props.nft.id)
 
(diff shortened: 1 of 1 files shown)


apps/web/src/archetypes/NFT/Modal/NftDialog.tsx
```diff
@@ -39,6 +39,8 @@ const backdropKeyframes = keyframes`
 const NftDialog = (props: NftDialogProps) => {
   const { nft, loading } = useNftById(props.nft.id)
 
+  console.log(nft)
```
tien, line 42: Prolly should remove this

tien (changes requested):

JAhimaz (commented):
Done, sigh silly mistake.

tien (approved):
lgtm

PR #409, opened 2022-11-25
fix: temporarily switched to v0a tx history indexer

alecdwm:
The v0b indexer is currently offline (ran out of disk space - we're waiting on a request to add more), so we've spun up another one (v0a) which is only indexing the past month-ish of moonbeam.


helloitsbirdo (approved):
lgtm

PR #412, opened 2022-11-27
fix: allow pjs and subwallet users to connect even if Talisman is not installed

0xKheops:
Fixed our Connect/Install button to allow user to connect with their existing wallet if they have one.

It was redirecting users to the talisman install page if it wasn't installed, preventing users to connect with polkadotjs or subwallet
It will now only redirect if there is no wallet installed, and show the existing wallet select modal otherwise

Note : this is a quick-fix solution, only 3 wallets are currently supported.


tien (approved):
lgtm

PR #413, opened 2022-11-28
chore: removed .devcontainer

alecdwm:
I added it and I'm fairly sure I was the only person using it.

I've since made a lot of customisations which are super specific to just me, so since I'm not maintaining the one in the repo anymore I figured it's better to remove the deprecated code.


tien (approved):
lgtm

PR #445, opened 2023-01-16
fix: incorrect portfolio value

### LINKED ISSUE #439: Portfolio values are incorrect
When I'm looking at my portfolio in the web-app, the `Portfolio value` (total) at the top is not reflective of my total assets + staking + crowdloans.

1. The fiat value of my free/available assets is not correct
2. Whether or not we want to include crowdloans as part of this or not is something else to consider, but staking should definitely be included.

Screen Shot 2023-01-05 at 3 26 46 pm

tien:
Fix #439


tien:
@JAhimaz once merged this change should also be carried over to the portfolio v2 branch

SwamiTLSM (approved):

PR #454, opened 2023-01-27
feat: added Banxa as the new buy option. (V1)

JAhimaz:
This is the V1 version which just redirects the user to the banxa url.

To meet the requirements of this pitch. UI that allows the user to select the amount, type of token and address will be added in the next version.


tien (approved):
lgtm

PR #456, opened 2023-01-30
fix: fixed logos for assets in breakdown and undefined

JAhimaz:
Some asset logos were undefined and asset breakdown didn't show proper logos. Temporary fix for evm is the eth token from coingecko, but fix for substrate related tokens is taken from the github.


tien (approved):
Lgtm

jonathanpdunne (approved):

PR #458, opened 2023-01-30
fix: fixed ajuna, shiden, amplitude and oak-network missing assets

JAhimaz:


0xKheops (approved):
LGTM, they now all display fine with my wallet

PR #459, opened 2023-01-30
fix: remove lock duration from claim rewards dialog

tien:
The Figma design is wrong, as claim rewards doesn't occur in any lock duration.


JAhimaz (approved):
lgmt 🚀

PR #462, opened 2023-01-31
fix: temporary added a loading to index.html

JAhimaz:
This is only a temporary solution to remove the jarring white screen.


tien:
Actual fix is in #467

PR #463, opened 2023-01-31
fix: raised cookie banner z-index

JAhimaz:
Cookie banner was under the welcome modal. Raised it's z-index to be higher than that. 🍪


tien (approved):
lgtm

PR #481, opened 2023-02-07
chore: improve development menu

tien:
- Fix dev menu styling & layout
- Implement code splitting to only serve development bundle on request (save on production bundle size)


jonathanpdunne (approved):
LGTM

PR #487, opened 2023-02-09
fix: moved airtable

JAhimaz:


tien:
@JAhimaz Can you also add the type information to `react-app-env.d.ts`

tien (approved):
LGTM

PR #502, opened 2023-02-17
chore: add crowdloans, fix copy

jonathanpdunne:
- Fix copy for no crowdloans
- Add Hashed crowdloan
- Add Shiden crowdloan

### DIFF
--- a/apps/web/src/archetypes/Crowdloan/Indexx.tsx
+++ b/apps/web/src/archetypes/Crowdloan/Indexx.tsx
@@ -99,9 +99,11 @@ const Index = styled(({ withFilter, className }: { withFilter: boolean; classNam
       <Await until={!loading}>
         <NoResults
           require={count?.filtered > 0}
-          title={t('noResult.title')}
-          subtitle={t('noResult.subtitle')}
-          text={t('noResult.text')}
+          // title={t('noResult.title')}
+          // subtitle={t('noResult.subtitle')}
+          // text={t('noResult.text')}
+          subtitle={t('noCrowdloans.text')}
+          text={t('noCrowdloans.subtext')}
         >
           <Grid>
             {crowdloans.map(({ id }) => (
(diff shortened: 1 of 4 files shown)


apps/web/src/archetypes/Crowdloan/Indexx.tsx
```diff
@@ -99,9 +99,11 @@ const Index = styled(({ withFilter, className }: { withFilter: boolean; classNam
       <Await until={!loading}>
         <NoResults
           require={count?.filtered > 0}
-          title={t('noResult.title')}
-          subtitle={t('noResult.subtitle')}
-          text={t('noResult.text')}
+          // title={t('noResult.title')}
+          // subtitle={t('noResult.subtitle')}
+          // text={t('noResult.text')}
```
tien, line 104: Can these be deleted rather than comment out?

apps/web/src/libs/talisman/util/_config.ts
```diff
@@ -1940,46 +1992,48 @@ export const crowdloanDetails: CrowdloanDetails[] = [
...
+  //     ],
+  //     custom: [
+  //       {
+  //         title: 'Auction Total',
+  //         value: '350,000,000 tokens',
+  //       },
+  //     ],
+  //     info: null,
+  //   },
+  // },
```
tien, line 2036: Same here, better delete this, the config file is already big as is

tien (changes requested):
Just need some clean up

apps/web/src/libs/talisman/util/_config.ts
jonathanpdunne: Yeah, I'm not super happy with this either. Opened [an issue re. this, plus other thoughts](https://github.com/TalismanSociety/talisman-web/issues/503)

jonathanpdunne: Removed

tien (approved):
lgtm

PR #506, opened 2023-02-22
fix: token tooltips

JAhimaz:
Todo

- Add Network to the tooltips


apps/web/src/components/recipes/Asset/Asset.tsx
```diff
@@ -193,6 +193,8 @@ const Asset = Object.assign((props: AssetProps) => {
   const { token, lockedAsset } = props
   const navigate = useNavigate()
 
+  console.log(token)
```
tien, line 196: Can this be removed?

PR #523, opened 2023-03-02
perf: bump balances library to version with performance fixes

tien:


0xKheops:
need to update deps again, pushed another fix on library side
it will prevent websockets from beeing recreated (and throwing errors) in case subscription has to be updated (user adds or removes an account, chain data receives an update from subsquid, etc.)

alecdwm:
Latest version (which @0xKheops is referring to) is [0.0.0-pr587-20230302015709](https://www.npmjs.com/package/@talismn/balances/v/0.0.0-pr587-20230302015709)

0xKheops (approved):
LGTM

PR #524, opened 2023-03-02
refactor: migrated explore page to baserow

JAhimaz:


apps/web/src/archetypes/Explore/hooks.tsx
```diff
@@ -13,31 +13,41 @@ export const useFetchDapps = () => {
...
-                  console.log('data err - ' + item.fields.name)
+                if (!item.name || !item.url) {
+                  console.log('data err - ' + item.name)
                   return undefined
                 }
 
-                if (!item.fields.logo) {
-                  console.log('logo err - ' + item.fields.name)
+                if (!item.logo?.[0]?.url) {
+                  console.log('logo err - ' + item.name)
```
tien, line 33: Can we remove the console logs? Or at least send them to sentry instead of logging?

apps/web/src/archetypes/Explore/hooks.tsx
```diff
@@ -13,31 +13,41 @@ export const useFetchDapps = () => {
...
                   return undefined
                 }
 
+                // take all the tags.value and put them in an array
+                if (item.tags) {
+                  item.tags = item.tags.map((tag: any) => tag.value)
+                }
```
tien, line 40: Should do this instead

```ts
return {
  // ...
  tags: item.tags?.map((tag: any) => tag.value) ?? [],
  // ...
}
```

apps/web/src/archetypes/Explore/hooks.tsx
JAhimaz: Dont think its worth sentrying them as everytime someone goes on the page it wont load due to the table not having the data. Best I'd just remove em.

apps/web/src/react-app-env.d.ts
```diff
@@ -18,7 +18,7 @@ declare namespace NodeJS {
     REACT_APP_SENTRY_RELEASE?: string
     REACT_APP_POSTHOG_AUTH_TOKEN: string
     REACT_APP_TX_HISTORY_INDEXER: string
-    REACT_APP_AIR_TABLE_API_KEY: string
+    REACT_APP_BASEROW_AUTH_TOKEN: string
```
tien, line 21: Can you also add an empty value to .env.example?

tien (approved):
lgtm

PR #545, opened 2023-03-13
feat: added baserow crowdloans

JAhimaz:


apps/web/src/archetypes/Parachain/Tags.tsx
```diff
@@ -1,19 +1,19 @@
...
-      {tags.map(tag => (
+      {/* {tags.map(tag => (
         <Pill key={tag} primary small>
           {tag}
         </Pill>
-      ))}
+      ))} */}
     </div>
   )
 })`
```
tien, line 19: Can these be deleted instead of commented out?

apps/web/src/libs/@talisman-crowdloans/provider.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+        // Skip iteration if either title or value is null or undefined
+        if (!crowdloan[titleKey] || !crowdloan[valueKey]) {
+          continue
+        }
+        customRewards.push({
+          title: crowdloan[titleKey],
+          value: crowdloan[valueKey],
+        })
+      }
```
tien, line 72: Can we remove all the imperitive for loops?

JAhimaz:
```suggestion
      const links: { [key: string]: string } = Object.keys(crowdloan).reduce(
        (acc: Record<string, string>, key: string) => {
          if (key.startsWith('links.')) {
            const value = crowdloan[key]
            if (value) {
              const newKey = key
                .replace(/^links\./, '')
                .replace(/-(\w)/g, (_, c) => c.toUpperCase())
                .toLowerCase()
              acc[newKey] = value
            }
          }
          return acc
        },
        {}
      )

      const customRewards = Object.keys(crowdloan)
        .filter(key => key.match(/^rewards\.custom\.\d+\.title$/))
        .filter(titleKey => {
          const valueKey = titleKey.replace('title', 'value')
          return crowdloan[valueKey] !== undefined
        })
        .map(titleKey => {
          const valueKey = titleKey.replace('title', 'value')
          return {
            title: crowdloan[titleKey],
            value: crowdloan[valueKey],
          }
        })
        .filter(reward => reward.title && reward.value)
```

tien (approved):
lgtm

PR #550, opened 2023-03-16
Use themeColors from chaindata

alecdwm:


tien:
Have incorporated these changes in ui overhaul PR

PR #566, opened 2023-03-22
Fix: (serif font) && (baserow crash) && (typescript sdk path)

alecdwm:
Fixes:
- serif font in UI when fonts still loading
- invalid baserow key crashing the UI
- vscode broken intellisense on unix due to strange `typescript.tsdk` path
  now using the one recommended [in the vscode docs](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript) instead


tien (approved):
Lgtm

PR #567, opened 2023-03-24
refactor: api read & instantiation

### LINKED ISSUE #342: Refactor `useChainState` & `useQueryMulti` hook
# Problem

Currently, [useChainState](src/domains/common/hooks/useChainState.ts) & [useQueryMulti](src/domains/common/hooks/useQueryMulti.ts) hook keep and handle storage subscription internally within the hook lifecycle.

This requires us to strategically identify the common component ancestor for each component that use storage subscription and have it there instead. As to not create too many active subscriptions.

This is ok in most cases, since generally you want to keep logic higher up in the components tree.

But this restriction hinders us from easily creating reusable hooks that depend on storage subscription without having them aware of the outer context:

```tsx
const useFoo = () => {
  // Create subscription on mount
  const loadableState = useChainState("query", "nominationPools", "minJoinBond");
  // ...
}

const useBar = () => {
  // Create subscription on mount
  const loadableState = useChainState("query", "nominationPools", "minJoinBond");
  // ...
}

const Talisman = () => {
  // 2 duplicate subscriptions are created 
  const foo = useFoo();
  const bar = useBar();
}
```

# Ideal solution

- Implement centralised chain cache layer for chain subscription
- Refactor [useChainState](src/domains/common/hooks/useChainState.ts) & [useQueryMulti](src/domains/common/hooks/useQueryMulti.ts) to be simple cache look up with key `{type}:{module}:{section}:{params}`

```tsx
const useFoo = () => {
  // Get called 1st, no cache exist. Establish subscription with cache key `query:nominationPools:minJoinBond`
  const loadableState = useChainState("query", "nominationPools", "minJoinBond");
  // ...
}

const useBar = () => {
  // Get called 2nd, cache exist with key `query:nominationPools:minJoinBond`. Subscribe to the storage value
  const loadableState = useChainState("query", "nominationPools", "minJoinBond");
  // ...
}

const Talisman = () => {
  const foo = useFoo();
  const bar = useBar();

  useEffect(() => {
    return () => {
      // Component unmount. No active subscriber of key `query:nominationPools:minJoinBond`. Tear down subscription
    };
  });
}
```

Also, by having subscription in a cache layer outside of React render context. We can implement support for react suspense:

```ts
const useChainState = () => {
  const cache = readSomeCache();

  if (cache.content instanceof Error) {
    // To be catch by React Error Boundary
    throw cache.content;
  }

  if (cache.content instanceof Promise) {
    // Suspense rendering
    throw cache.content;
  }

  return cache.content;
}
```

tien:
- cache api subscription in recoil atoms
- cache `ApiPromise` in atom with tear down
- implement `Suspense` first api

Close #342


jonathanpdunne (approved):
Awesome

PR #599, opened 2023-05-05
fix: use talisman wallet rpcs for balance requests when available

alecdwm:
This PR updates the balances libs to a version which will make use of the wallet injected `chain-connector` when one is available.

The outcome of this is that Talisman users will no longer encounter browser websocket limits when viewing both the wallet and portal at the same time.

Future work: expose the injected `chain-connector` to the remainder of the portal via recoil - then rework existing rpc connections in the portal (e.g. in the staking, crowdloans, etc features) to make use of said `chain-connector` for their rpc requests so that all substrate rpc requests share the one websocket connection across portal and wallet.

More future work: refactor the glue code between the output of the balances libs and the portfolio view in the portal so that it can show all of the information which is available in the portfolio view of the wallet.


tien:
Lit 🔥

tien (approved):
LGTM 💯

PR #617, opened 2023-05-15
feat: fast unstake

tien:
- [ ] Check performance on mobile. cc @replghost 
- [ ] Final check of criteria


tien:
@jonathanpdunne Final check, does the user need to not have received any rewards in the last 28 era/days (for Polkadot) including the current era. Or 28 era/days + the current era, for a total of 29?

replghost:
I think you mean 2 rather than 28?

tien:
> I think you mean 2 rather than 28?
> 
> Since payouts happen once per era/calendar day, my guess would be if the last two payments didn't happen.

https://support.polkadot.network/support/solutions/articles/65000182201-staking-dashboard-how-to-unbond-your-tokens

it's basically for the full unbonding period

jonathanpdunne (approved):

Issue #663, opened 2023-06-01
[Webapp] DEX Meta Aggregator


0byt:

0byt:
@chidg integration portion delayed 1 day due to Root account refactor, he will talk to @tien to see if it's more efficient to handover this part to tien instead

0byt:
FAQ sheet here (to be edited) @jonathanpdunne 
https://docs.google.com/document/d/1rOGGSqf1n10SHtevAkBZdhdMT_1ZsOD5b98GInPYiOk/edit

tien:
Need to evaluate Privadex further

PR #664, opened 2023-06-01
refactor: use new native CSS `color-mix` functionality

tien:
For the "Orb" component that can't be migrated. Only import needed functions for tree-shaking.


tien:
Part of #663

PR #695, opened 2023-06-28
[DO NOT MERGE] refactor: text component color

tien:
Use colour from CSS rather than JS props.

Requires fix from https://chromium-review.googlesource.com/c/chromium/src/+/4601325


tien:
Superseded by #1009

PR #709, opened 2023-07-05
chore: re-enable tx history with notice message

tien:


chidg (commented):
The text is a bit small - suggest to make it a bit larger, and maybe in some kind of banner or alert UI element to make it stand out more.

tien:
> The text is a bit small - suggest to make it a bit larger, and maybe in some kind of banner or alert UI element to make it stand out more.

We don't have an alert element in the UI lib atm. But would something like this be sufficient?

image

chidg:
> We don't have an alert element in the UI lib atm. But would something like this be sufficient?

 LGTM!

PR #784, opened 2023-08-30
refactor: upgrade floating UI

tien:
- Bump library version
- Migrate to using `refs` object
- Use new `useClientPoint` instead of custom hook
- For tooltip, use container ref with `display: 'contents'` to avoid having to pass floating props


lee-qa:
@tien regression tested on the tooltips and LGTM. let me know if need testing for other areas.

PR #790, opened 2023-09-05
fix: use multisig id as identifier

chrisling-dev:
This PR updates Transactions to use their Multisig ID, which is constructed from [Timepoint](https://paritytech.github.io/substrate/master/pallet_multisig/struct.Timepoint.html), instead of call hash as unique identifier. Call hashes aren't necessarily unique because transactions that do the exact same thing will have the same hash value.

Before:
Screenshot 2023-09-06 at 1 30 38 AM

After:
Screenshot 2023-09-06 at 1 31 09 AM


chrisling-dev:
> Looks good! This might be a little opinionated, but I would name this as `transactionID`/ `callID` / or similar, to save possible confusion between this and the concept of the identifier for a multisig wallet.

Sure! Named this as Multisig ID because that's what they have on subscan UI. `transactionID` works too!

PR #794, opened 2023-09-07
feat(signet): form to select a proposal and enter vote parameters

chrisling-dev:
This PR creates a UI for users to easily vote on a proposal as a Multisig.

Create and Approve a Vote tx:
https://github.com/TalismanSociety/talisman-web/assets/81092286/d5636b23-18d3-4d3e-bd7c-9e229022c4ea

View, Approve and Execute a Vote tx:
https://github.com/TalismanSociety/talisman-web/assets/81092286/dbe793f4-ada9-4b69-891a-bb98dccb3141

### DIFF
--- a/apps/multisig/src/domains/chains/tokens.ts
+++ b/apps/multisig/src/domains/chains/tokens.ts
@@ -203,6 +203,7 @@ export type Chain = {
   rpcs: Rpc[]
   ss58Prefix: number
   subscanUrl: string
+  polkaAssemblyUrl?: string
 }
 
 export const chainTokensByIdQuery = graphQLSelectorFamily({
--- a/apps/multisig/src/layouts/Overview/NewTransactionModal/vote/index.tsx
+++ b/apps/multisig/src/layouts/Overview/NewTransactionModal/vote/index.tsx
@@ -1,6 +1,12 @@
 import { css } from '@emotion/css'
 import { useMemo, useState } from 'react'
-import { VoteDetails, isVoteFeatureSupported } from '@domains/referenda'
+import {
+  SplitAbstainVoteParams,
+  StandardVoteParams,
+  VoteDetails,
+  isVoteDetailsComplete,
+  isVoteFeatureSupported,
+} from '@domains/referenda'
 import VotingForm from './VotingForm'
 import { useRecoilValue, useRecoilValueLoadable } from 'recoil'
 import {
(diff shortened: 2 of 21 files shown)


apps/multisig/src/layouts/Overview/NewTransactionModal/vote/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+        padding: 32px;
+      `}
+    >
+      <VotingForm
+        voteDetails={voteDetails}
+        onChange={setVoteDetails}
+        onCancel={onCancel}
+        onNext={() => setReviewing(true)}
+      />
+      <SideSheet open={reviewing} onRequestDismiss={() => setReviewing(false)} title="" />
```
chrisling-dev, line 36: Completing this in next PR to keep this PR small and reviewable

apps/multisig/src/layouts/Overview/NewTransactionModal/vote/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+        balance: new BN(0),
+        vote: {
+          conviction: 1,
+          aye: true,
+        },
+      },
+    },
+  })
+
+  // instead of allowing the user to select any token later on, we just use the first native token of the chain
```
chrisling-dev, line 50: Not sure if this is actually true across every chains tho, are there any chains that allow voting with a secondary token?

However even if there is, it won't be high priority to fix this now as we're only supporting some of the popular chains

apps/multisig/src/layouts/Overview/NewTransactionModal/vote/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+    apiLoadable.contents.tx.convictionVoting,
+    apiLoadable.contents.tx.proxy,
+    apiLoadable.state,
+    isPalletSupported,
+    multisig.proxyAddress.bytes,
+    nativeToken,
+    voteDetails,
+  ])
+
+  const transactionName = useMemo(() => {
```
chrisling-dev, line 86: leaving this in a useMemo as it will get more complex as we introduce Split and Abstain

chrisling-dev (commented):
Self reviewed

apps/multisig/src/domains/chains/tokens.ts
```diff
@@ -203,6 +203,7 @@ export type Chain = {
   rpcs: Rpc[]
   ss58Prefix: number
   subscanUrl: string
```
jonathanpdunne, line 205: should this be an optional attribute as well?

jonathanpdunne (approved):

apps/multisig/src/domains/chains/tokens.ts
chrisling-dev: is it common for substrate chain to not have a subscan url?

I think can keep as is anyway since all the chains we support have subscanUrl. The polka assembly url has to be optional because the testnet chains dont support the opengov pallets and hence dont have a polka assembly url

PR #797, opened 2023-09-11
refactor: multi-currency improvements

tien:
- stronger type checking
- fix intermittent bug where animated number might reset to 0 when currency change


lee-qa (commented):
Regression tested and looks good

PR #819, opened 2023-09-27
Support Hasura validated insert action

liamaharon:
- Supports changes made in https://github.com/TalismanSociety/signet-self-hosted-metadata-backend/pull/3
- Adds proper attribution to p.js for the `decodeCallData` function
- Adds error toast if insert fails


chrisling-dev (approved):
LGTM, will merge shortly

PR #823, opened 2023-10-05
feat: signet siws

chrisling-dev:
Changes:
- Introduce a minimal version of SIWS to the frontend, which is used to generate a JWT token to authenticate user for data from the backend
- all data from backend is protected and requires a JWT token
- Frontend tracks a list of `Teams` that are stored in the backend. Teams are then parsed into multisig on the frontend for backward compatibility
- In the future, Teams can represent different setups (e.g. proxy -> proxy instead of multisig -> proxy)
- Import page is deprecated: by querying the list of teams user is in from the backend, we essentially enable "auto-import", more magical than magic link
- how metadata is fetched is completely changed: 
  - the previous approach makes 1 query per transaction. This means if you're importing a vault that already has 100+ transactions, the app will make 100+ requests to the backend. It was also doing read + write in the middle of reading tx list, which was somewhat difficult to maintain.
  - the new approach fetches a list of protected tx metadata asynchronous to how pending tx list is being queried and stores them in a `map` (not exactly map, using Record). The tx list then simply consumes data from the tx list, which makes the code easier to read and make querying tx list more "light weight"
-  Create vault: The create page now doesn't care about "No Vault" step. The Layout component will check if user doesn't have a vault and if the requireMultisig flag is true, then it will show the No Vault step instead. This reduces the number of redirect, and makes it easier to share such "guard" logics with other pages

Remaining:
- handle updating team config after change
- auto sync multisig config from the backend (fe has to trigger a request tho)

Nice to have:
- simple ui to add previously created vaults to backend?


apps/multisig/src/domains/chains/extrinsics.ts
```diff
@@ -410,6 +411,9 @@ export const useApproveAsMulti = (
       onFailure: (message: string) => void
       metadata?: TxOffchainMetadata
     }) => {
+      // cache selected account when tx was approved
+      const signedInAs = signedInAccount
```
chrisling-dev, line 415: to prevent failure in inserting a tx if user creates and approve tx -> turn off modal -> switch acc to do something else which the new acc does not have access to the prev vault

apps/multisig/src/domains/extension/ExtensionWatcher.ts
```diff
@@ -0,0 +1,73 @@
+import { useCallback, useEffect, useState } from 'react'
+import { useRecoilState, useSetRecoilState } from 'recoil'
+import { accountsState, extensionAllowedState, extensionLoadingState } from './index'
+import { web3AccountsSubscribe, web3Enable } from '@polkadot/extension-dapp'
+import toast from 'react-hot-toast'
+import { uniqBy } from 'lodash'
+import { Address } from '@util/addresses'
+
+export const ExtensionWatcher = () => {
```
chrisling-dev, line 9: simply moved from here https://github.com/TalismanSociety/talisman-web/pull/823/files#diff-0764a0f4ae31213f6ed7a7539564aee35bd129b48ec5d3b33a0f5ab601dd931aL34

apps/multisig/src/layouts/Overview/Transactions/index.tsx
```diff
@@ -159,6 +159,13 @@ const TransactionsList = ({ transactions }: { transactions: Transaction[] }) =>
                               multisig.proxyAddress,
                               expectedNewMultisigAddress
                             )
+                            // rather than doing this in the backend, each team has a 'resync' function
+                            // 1. that makes a backend call
+                            // 2. backend call will find the latest config change
+                            // 3. compute the new multisig address
+                            // 4. backend test if the multisig address is proxy of the proxied account
+                            // 5. if it is, update the multisig config and delagatee addres
+                            // 6. else mark this as a bad team (maybe show a page where they can correct it)
```
chrisling-dev, line 168: notes for myself, ignore for now

apps/multisig/src/layouts/CreateMultisig/Confirmation.tsx
```diff
@@ -168,78 +162,23 @@ const Confirmation = (props: {
...
-            }
-          `}
-        >
-          <h2
-            className={css`
-              font-size: 20px;
-              margin-bottom: 8px;
-            `}
-          >
-            Depositor
```
chrisling-dev, line 200: Deposit settings is removed because now the depositor is always the selected account in header

chrisling-dev (commented):
self reviewed

apps/multisig/src/domains/auth/index.ts
```diff
@@ -0,0 +1,142 @@
+import { useCallback, useState } from 'react'
+import { atom, selector, useRecoilState, useSetRecoilState } from 'recoil'
+import { web3FromSource } from '@polkadot/extension-dapp'
+import { InjectedAccount, accountsState } from '../extension'
+import persistAtom from '../persist'
+import toast from 'react-hot-toast'
+
+const SIWS_ENDPOINT = process.env.REACT_APP_SIWS_ENDPOINT ?? ''
```
chrisling-dev, line 8: initial approach was to make siws endpoints as hasura actions so the frontend only deals with 1 endpoint for all backend needs, but changed to using a separate endpoint for SIWS for 2 reasons:
- SIWS requires maintaining sessions between 2 network requests, and using Hasura actions will cause the sessions to break
- In the future we may abstract SIWS completely from our stack to a separate service

apps/multisig/src/domains/extension/ExtensionWatcher.ts
```diff
@@ -0,0 +1,96 @@
...
+      // initial connections made
+      const connectedExtensions = uniqBy(accounts, account => account.meta.source).map(account => account.meta.source)
+
+      // browsers like Arc will inject each extensions synchronously
+      // AccountWatcher requires all extensions to be injected to hydrate UI properly
+      if (!extensionInitiated) {
+        if (detectedExtensions.every(extension => connectedExtensions.includes(extension))) {
+          setExtensionInitiated(true)
+        }
+      }
```
chrisling-dev, line 75: and added this logic

apps/multisig/src/domains/tx-history/index.tsx
```diff
@@ -164,50 +159,27 @@ export const useConfirmedTransactions = () => {
...
+                  `Failed to create extrinsic from callData recieved from metadata sharing service for transactionID ${transactionID}`
+                )
+              }
+              const derivedHash = extrinsicDerivedFromMetadataService.registry
+                .hash(extrinsicDerivedFromMetadataService.method.toU8a())
+                .toHex()
+              if (derivedHash !== hash) {
+                throw new Error(
+                  `CallData from metadata sharing service for transactionID ${transactionID} does not match hash from chain. Expected ${hash}, got ${derivedHash}`
+                )
```
chrisling-dev, line 176: Goal of changes in this file is to:
- Remove code that requests for data from metadata service on every transactions
- Instead, try to find the metadata from `txMetadataByTeamId` which sends very minimal network requests to our backend

chrisling-dev (commented):
self reviewed

PR #851, opened 2023-11-13
fix: temporarily remove kusama XCM routes

jonathanpdunne:
Disable Kusama XCM routes, due to an issue with Kusama XCM teleports and RATs. Example:
https://kusama.subscan.io/xcm_message/kusama-98082ccbd5ae3e416b17276a0aaaeadd85aecb7a


tien (approved):
lgtm

PR #868, opened 2023-12-06
fix: only show balances from selected account

tien:


jonathanpdunne (approved):
Small thing: maybe separate css changes to AccountSelector into a different PR in the future. Otherwise LGTM.

PR #898, opened 2023-12-13
chore: disable SLPx staking

tien:


tien:
No longer needed with fix from #899

PR #942, opened 2024-01-23
refactor: allow passing of tx builder function to `useExtrinsic` hook

tien:
# Example

```ts
const extrinsic = useExtrinsic(
  useCallback(
    (api: ApiPromise) => api.tx.utility.batchAll([api.tx.remark.store('foo'), api.tx.remark.store('bar')]),
    []
  )
)
```


tien:
@alecdwm could be useful for some of the Crowdloans stuff

PR #954, opened 2024-01-29
feat: integrate with signet sdk

chrisling-dev:


tien:
lgtm, @chrisling-dev There's also `useExtrinsicBatch` which is deprecated & used in one place only. Can you either add signet support to that or just migrate the 1 usage over to `useExtrinsic` instead?

apps/web/src/domains/accounts/recoils.ts
```diff
@@ -127,6 +129,9 @@ export const wagmiAccountsState = atom<Account[]>({
 export const accountsState = selector({
   key: 'Accounts',
   get: ({ get }) => {
+    const signetAccount = get(signetAccountState)
+    if (signetAccount) return [{ ...signetAccount, readonly: false, partOfPortfolio: true }]
```
tien, line 133: Just nit-picking but would be better to do explicit check like `signetAccount !== undefined` if possible.

apps/web/src/domains/accounts/recoils.ts
```diff
@@ -264,3 +272,32 @@ export const AccountWatcher = () => {
 
   return null
 }
+
+export const signetAccountState = atom<InjectedAccount>({
+  key: 'SignetAccountState',
```
tien, line 277: The key here should be `SignetAccount`

apps/web/src/domains/accounts/recoils.ts
```diff
@@ -264,3 +272,32 @@ export const AccountWatcher = () => {
...
+  default: undefined,
+})
+
+export const SignetWatcher = () => {
+  const { inSignet, sdk } = useSignetSdk()
+  const setSignetVaultAccount = useSetRecoilState(signetAccountState)
+
+  useEffect(() => {
+    if (inSignet && sdk) {
```
tien, line 286: Do explicit `undefined` check on `sdk` if possible

apps/web/src/domains/common/hooks/useExtrinsic.ts
```diff
@@ -100,6 +105,12 @@ export function useExtrinsic(
           })
 
           try {
+            if (signingWithSignet && submittable) {
+              const { ok, error, receipt } = await sdk.send(submittable.method.toHex())
+              // this doesnt matter because signet will redirect users away after every transaction
+              if (ok && receipt) reject(new Error('Please ignore this message.'))
+              if (error) reject(new Error(error))
```
tien, line 112: Probably better to do `new Error('Some descriptive msg here', { cause: error })`

tien (changes requested):
Just couple of minor points, also the previous msg about `useExtrinsicBatch`, everything else look good 💪

apps/web/src/domains/accounts/recoils.ts
chrisling-dev: ahh np, using pattern from signet lol

apps/web/src/domains/accounts/recoils.ts
chrisling-dev: I removed this check because sdk cannot be undefined in the latest version

chrisling-dev:
> lgtm, @chrisling-dev There's also `useExtrinsicBatch` which is deprecated & used in one place only. Can you either add signet support to that or just migrate the 1 usage over to `useExtrinsic` instead?

Not sure what's the best way to migrate yet given useExtrinsic doesnt return the extrinsic for batching, but instead just exposes the signAndSend function, I assume to migrate you need to build both 'staking.chill' and 'staking.unbond' as extrinsics and then batch them? 

So seems like it needs to be
```tsx
const batchExtrinsic = useExtrinsic('utility', 'batch')
...

return {
  ...
  unbondAll: async (account: string) => {
    // const chillExtrinsic = ... build chill extrinsic
    // const unbondExtrinsic = ...build unbond extrinsic
    return batchExtrinsic.signAndSend(account, {
    //  ... pass in chillExtrinsic and unbondExtrinsic
    })
  }
}
```

Lmk if you have something else in mind or I can migrate according to the snippet above

tien:
@chrisling-dev if you check the function signature of `useExtrinsic`, it also accepts either a submittable or a function that generate a submittable, so you can do something like this:

```ts
const extrinsic = useExtrinsic(useCallback((api) => api.tx.utility.batchAll([
  // ... stuffs goes here
]), []))
```

chrisling-dev:
> @chrisling-dev if you check the function signature of `useExtrinsic`, it also accepts either a submittable or a function that generate a submittable, so you can do something like this:
> 
> ```ts
> const extrinsic = useExtrinsic(useCallback((api) => api.tx.utility.batchAll([
>   // ... stuffs goes here
> ]), []))
> ```

Migrated and removed useExtrinsicBatch! Lmk wyt

apps/web/src/domains/common/hooks/useExtrinsic.ts
```diff
@@ -105,7 +105,7 @@ export function useExtrinsic(
           }
         })()
 
-        const signingWithSignet = signetAccount.address === account
+        const signingWithSignet = signetAccount !== undefined && signetAccount.address === account
```
tien, line 108: Maybe this can be shortened to: `signetAccount?.address === account`

tien (approved):
lgtm

apps/web/src/domains/common/hooks/useExtrinsic.ts
chrisling-dev: yeap np, addressed

PR #973, opened 2024-02-09
feat: kodadot NFTs

alecdwm:
- Fixes kodadot-minted NFTs
- Adds `(Polkadot|Kusama)-Asset-Hub` NFTs


tien (changes requested):
Basically lgtm, just a couple of things:

- Look like 2 of the endpoint have identical schema, so they can probably share the same codegen output
- In saying that, why is the `RMRK` one has different schema? Is it because it's on V2?

tien (approved):
lgtm

vikiival:
> In saying that, why is the RMRK one has different schema? Is it because it's on V2?

RMRK2 has slightly different format as we need to accommodate rmrk2 specific thingies. 


But 80% should be the same
https://www.youtube.com/watch?v=Egef-NfGGfI

PR #974, opened 2024-02-09
feat: enable Astar DApp staking

tien:


tien:
Should only be merge after DApp staking v3 is live on Astar

PR #1058, opened 2024-03-18
feat: on-chain-id

alecdwm:


tien:
For the on-chain-id lib thing, can we use camel instead of kebab case?

apps/web/src/libs/on-chain-id/hooks/useResolveNsName.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  return [address, { isNsLookup, nsLookupType, isNsFetching }] as const
+}
+
+const cacheKey = 'TalismanNsNamesCache'
+const persistItemDuration = 15_778_476_000 // 6 months in milliseconds
+const nsNamesCache = new Map<string, { result?: [string, NsLookupType] | null; updated?: number }>(
+  JSON.parse(localStorage.getItem(cacheKey) ?? '[]')
+)
+const persistNsNamesCache = () =>
```
tien, line 71: Is it slow enough to require caching? I rmb ENS resolution was almost instant.

apps/web/src/components/widgets/AccountsManagementMenu.tsx
```diff
@@ -156,7 +177,21 @@ const AccountsManagementMenu = (props: { button: ReactNode }) => {
...
+                    }
+                    &:hover .show-on-menu-item-hover {
+                      display: block;
+                    }
+                    &:hover .hide-on-menu-item-hover {
+                      display: none;
+                    }
+                  `}
+                  onClick={() => setSelectedAccountAddresses(() => [x.address])}
+                >
```
tien, line 194: This seems rather hacky and require knowing the internal implementation of `AccountsManagementAddress`.

Maybe we can change `MenuItem` `children` to optionally accept a render function, so you can do something like this:
```tsx
<Menu.Item>
  {({ hover }) => <AccountsManagementAddress name={x.name} address={x.address} ens={!hover} />}
</Menu.Item>
```

apps/web/src/libs/on-chain-id/hooks/useResolveNsName.ts
alecdwm: Yeah it's necessary

apps/web/src/components/widgets/AccountsManagementMenu.tsx
```diff
@@ -64,6 +65,27 @@ const AccountsManagementSurfaceIconButton = (props: { size?: number | string })
...
+  name?: string
+  address: string
+  showOnChainId?: boolean
+}) => {
+  const [onChainId] = useOnChainId(address)
+
+  return (
+    <>
+      {onChainId && (
+        <div css={{ display: showOnChainId ? 'block' : 'none', color: 'var(--color-primary)' }}>{onChainId}</div>
```
tien, line 82: `var(--color-primary)` can be changed to `theme.color.primary`

tien:
There are a couple of `null` being used, I think it's better to use `undefined` & avoid `null` unless absolutely necessary

apps/web/src/libs/onChainId/hooks/useOnChainId.ts
```diff
@@ -0,0 +1,47 @@
...
+    persistOnChainIdsCache()
+  }, [address, result, live])
+
+  const onChainId = live ?? cached ?? null
+
+  return [onChainId] as const
+}
+
+const cacheKey = 'TalismanOnChainIdsCache'
+const persistItemDuration = 15_778_476_000 // 6 months in milliseconds
```
tien, line 29: I'm a bit worry about the arbitrary chosen cache time of 6 months. Let discuss this tmr.

PR #1067, opened 2024-03-27
fix: incorrect Polkadot Asset Hub ED

tien:


jonathanpdunne (commented):
Why do we need to remove the USDC entry?

tien:
@jonathanpdunne this is part of upstream now last I checked

jonathanpdunne (approved):
Sweet. LGTM.

PR #1073, opened 2024-04-09
build: bump `polkadot.js` dependencies

tien:


apps/web/src/libs/crowdloans/useCrowdloanContribute.tsx
```diff
@@ -1303,7 +1303,7 @@ async function buildAcalaTx({
   }
 
   const txs = [
-    api.tx.balances.transfer(proxyAddress, amount),
+    api.tx.balances.transferKeepAlive(proxyAddress, amount),
```
tien, line 1306: @alecdwm latest version of `pjs` remove `transfer`. This is what we want to do instead now, right?

alecdwm: Yeah! This LGTM

PR #1098, opened 2024-04-23
feat: artzero nfts

alecdwm:


packages/nft/src/generators/artZero.ts
```diff
@@ -0,0 +1,137 @@
...
+    if ((result?.ret?.length ?? 0) < 1) break
+
+    const type = 'artzero' as const
+    const chain = 'aleph-zero' as const
+
+    yield* (result?.ret ?? []).map(
+      (nft, index): Nft<'artzero', 'aleph-zero'> => ({
+        type,
+        chain,
+        id: `${type}-${chain}-${nft._id ?? index + offset}`,
```
tien, line 105: Is the `_id` field optional?

alecdwm: I don’t trust the API

packages/nft/src/generators/artZero.ts
```diff
@@ -0,0 +1,137 @@
...
+      method: 'POST',
+      headers: {
+        Accept: '*/*',
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
+      body: new URLSearchParams(
+        Object.entries({ owner, limit, offset, sort, isActive }).map(([key, value]) => [key, value.toString()])
+      ).toString(),
+    })
+    if (!response.ok) throw new Error(`ArtZero api response not ok: ${response.status}`)
```
tien, line 92: Can we add the response object to the error `cause` instead?

packages/nft/src/generators/artZero.ts
tien: In that case, I think it's better to just throw an error rather than returning an item without it actual ID. Because the ID will potentially be stored in local storage if the user favourite or hide the NFT.

tien (approved):
lgtm

PR #1147, opened 2024-06-28
[FIX] - Astar Staking issues

UrbanWill:
# Description

## ⛑️ **What was done:**

This PR fixes the following issues:

1. Hide ellipsis menu when no operation is available
2. Attention required red dot being displayed on ellipsis menu top right when it should not be displayed
3. Not earning rewards message when tokens are actually staked and earning rewards
4. Unable to claim staking rewards

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Ready to merge

### 🗒️ **Notes:**

Was not able to reproduce item 3, may still exist. I suspect it could be a stale value, memoized `earningRewards` value calculation to mitigate the issue.

### DIFF
--- a/apps/portal/src/components/widgets/staking/dappStaking/Stakes.tsx
+++ b/apps/portal/src/components/widgets/staking/dappStaking/Stakes.tsx
@@ -83,7 +83,7 @@ const Stake = ({
           )
         }
         lockedButton={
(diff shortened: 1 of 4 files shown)


chidg (approved):
LGTM!

PR #1154, opened 2024-07-24
feat: improved swap amount

chrisling-dev:
- swap amount subtracts gas buffer when swapping from native token
- allow transferAllowDeath


UrbanWill (approved):
LGTM

PR #1159, opened 2024-08-02
fix: unique nfts

alecdwm:
Fixes:
- EVM accounts not fetching unique NFTs
- Unique NFT images not showing due to change in the API response


UrbanWill (commented):
LGTM

PR #1161, opened 2024-08-06
:technologist: hide RMRK2 NFTs

vikiival:
### Context

We would like to turn off RMRK 2 indexer for some period of time (update / maintenance)
and we would like to not to break talisman wallet ^-^ 

### Fix

remove import of `createRmrk2NftAsyncGenerator`.


vikiival:
cc @alecdwm

alecdwm (approved):
LGTM! Thank you

PR #1173, opened 2024-09-06
fix: septemberfest

0xKheops:


UrbanWill (approved):

apps/portal/src/domains/common/recoils/api.ts
```diff
@@ -8,21 +7,20 @@ import { selectorFamily, useRecoilCallback } from 'recoil'
...
-          provider: new WsProvider(endpoint),
-          types: AvailJsSdk.spec.types,
-          rpc: AvailJsSdk.spec.rpc,
-          signedExtensions: AvailJsSdk.spec.signedExtensions,
-        })
-
-      return await ApiPromise.create({ provider: new WsProvider(endpoint) })
-    },
+  // DO NOT USE any atom dependency here, nothing should invalidate an api object once created
+  get: endpoint => async () => {
```
chidg, line 11: Any reason to use the regex instead of the previous method based on known rpcs? Seems like it would be better not to rely on avail rpcs always matching this pattern.

0xKheops: yes, previous method required a dependency on the chains list atom, which changes multiple times as individual network configurations are fetched from github.
This lead to multiple api objects being created per chain, and was one of the problems solved by this PR.

0xKheops: i assume we could make it cleaner by providing an isAvail flag as part of the key for that atom family, though as mentioned during dev sync i think this should be rewritten without recoil and leverage Talisman Wallet's proxied RPCs. so i kept it as simple as possible.

PR #1183, opened 2024-09-30
fix: tao staking fee estimation

UrbanWill:
# Description

Fixes "Max" TAO staking fee estimation.

A dummy `feeEstimateTx` was being used for estimation the transaction fee, when actually `tx` was being submitted. The fix is using the real `tx` that is submitted to calculate the fees.

### 🖼️ **Screenshots:**

Note: After submitting the staked % will show 94%, this is due to the fact that other connected wallets also hold TAO in them. It has in fact staked 100% of the walled used for this transaction, I would have to wait another 360 blocks to record another demo.

https://github.com/user-attachments/assets/634f33eb-534f-4771-a06e-98cca7873b0e


chidg (approved):
LGTM!

PR #1187, opened 2024-10-11
Feat/lifi swap

chrisling-dev:


apps/portal/src/components/widgets/chainflip-swap/side-panel/SwapDetails.tsx
```diff
@@ -64,14 +66,28 @@ const Details: React.FC = () => {
...
+    }
+  }, [selectedProtocol, setSelectedProtocol, quotes, cachedQuotes, setSelectedSubProtocol])
+
+  useEffect(() => {
+    if ((!selectedSubProtocol || !selectedProtocol) && cachedQuotes.length > 0) {
+      const defaultQuote = cachedQuotes[0]
+      if (defaultQuote?.quote.state === 'hasData' && defaultQuote.quote.data?.subProtocol)
+        setSelectedSubProtocol(defaultQuote.quote.data?.subProtocol)
+    }
+  }, [cachedQuotes, selectedProtocol, selectedSubProtocol, setSelectedSubProtocol])
```
UrbanWill, line 90: All these useEffects have the same same dependencies, or are related to each other. Would be nice to boil them down to a single useEffect hook if possible, but that's a nitpick.

The following **_seems_** to work with some basic manual testing:

```
  useEffect(() => {
    // Cache quotes if they're available and all are in 'hasData' state
    if (quotes.state === 'hasData' && quotes.data?.length) {
      if (cachedQuotes.length === 0 || quotes.data.every(q => q.quote.state === 'hasData')) {
        setCachedQuotes(quotes.data)
      }
    }

    // Reset protocol selection if no valid protocol found in cached quotes
    if (!cachedQuotes.find(q => q.quote.state === 'hasData' && q.quote.data?.protocol === selectedProtocol)) {
      setSelectedProtocol(null)
      setSelectedSubProtocol(undefined)
    }

    // Automatically select a sub-protocol if none is selected and cached quotes are available
    if ((!selectedSubProtocol || !selectedProtocol) && cachedQuotes.length > 0) {
      const defaultQuote = cachedQuotes[0]
      if (defaultQuote?.quote.state === 'hasData' && defaultQuote.quote.data?.subProtocol) {
        setSelectedSubProtocol(defaultQuote.quote.data?.subProtocol)
      }
    }
  }, [quotes, cachedQuotes, selectedProtocol, selectedSubProtocol, setSelectedProtocol, setSelectedSubProtocol])
  ```

File: apps/portal/src/components/widgets/chainflip-swap/swap-modules/lifi.swap-module.ts [hunk 2]
```diff
@@ -0,0 +1,275 @@
...
+          id: t.address === zeroAddress ? evmNativeTokenId(id) : evmErc20TokenId(id, t.address),
+        }
+      })
+    })
+  return tokens.flat()
+})
+
+const routesAtom = atom(async get => {
+  try {
+    const fromAddress = get(fromAddressAtom) ?? '0x70045A9F59A354550EC0272f73AAe03B01Fb8a7a'
```
UrbanWill, line 62: Would be nice to have this contract address in a variable roughly named after what it is

apps/portal/src/components/widgets/chainflip-swap/swap-modules/lifi.swap-module.ts
```diff
@@ -0,0 +1,275 @@
...
+  protocol: PROTOCOL,
+  fromAssetsSelector,
+  toAssetsSelector: fromAssetsSelector,
+  quote: quoteAtom,
+  swap,
+  decentralisationScore: DECENTRALISATION_SCORE,
+  approvalAtom,
+}
+
+const retryStatus = async (
```
UrbanWill, line 249: My Jotai knowledge may be limited, but from what I have seen it doesn't offer a good fetch & retry solution of the box. Here is a suggestion for a future implementations:

Give `jotai-tanstack-query` a go, in this case instead of rolling our own retry logic, we could have used [atomWithQuery](https://jotai.org/docs/extensions/query#atomwithquery-usage) with `sdk.getStatus({ txHash: id })` as `queryFn` and leveraged react-query [Query Retries](https://tanstack.com/query/v5/docs/framework/react/guides/query-retries) to handle failed requests

UrbanWill (approved):

apps/portal/src/components/widgets/chainflip-swap/side-panel/SwapDetails.tsx
chrisling-dev: I think its better to separate useEffect hooks that serve different purpose, like managing cachedQuotes and selected protocol are 2 different things. But I can see that the last 2 hooks can definitely be combined

(also refactored the cachedQuotes hooks anyway to be 1 single hook)

PR #1199, opened 2024-10-21
chore: portal cleanup

alecdwm:
Included changes:
- Added `prettier-plugin-tailwindcss` for deterministic sorting of tailwind css classes
- Switched from `pretty-quick` (which was throwing errors with `prettier-plugin-tailwindcss`) to `lint-staged` (which we also use in the wallet repo)
- Added `@ianvs/prettier-plugin-sort-imports` for imports sorting (with the same config as we use in the wallet repo)
  - Unlike `@trivago/prettier-plugin-sort-imports`, this one is compatible with more recent versions of prettier
  - It also plays well with other prettier plugins, whereas the previous one would conflict with them
  - It also supports a better sorting order, which will now be automatically applied to files as we change them going on
- Updated the README because this project is now called Talisman Portal, not Talisman Web
- Deleted the old apps/portal/README.md which is a holdover from before this was a monorepo
- Updated the eslint config to consider `no-unused-vars` as a WARN, not an ERROR
  - In vscode this means `jump to next error` will find the next **useful** error as you're working on the code, and will only jump to an unused var once all actually important errors have been fixed
- Updated all favicons and logos to use the new Talisman logo 🪬
  - Also applied [2024 favicon best practises](https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs)
- Updated balances and chaindata libraries to use the newer `pub/v2` version of chaindata
- Updated a few helpful dependencies to the next non-breaking versions: `bignumber.js`, `tailwind-merge`, `tailwindcss`, `typescript`, `vite`, `mime`, `mime-db`
- Fixed Portal not connecting to wallets while the UI is suspended (by moving the wallet connection context provider out from the top-level Suspense)
- Fixed the account connection guard flashing up every time a page is loaded/reloaded (except for on the Portfolio, where it still flashes because I need to spend some more time debugging the behaviour of the portfolio Stakes overview)
  - The UI now suspends until an attempt has been made to connect to a wallet provider (or its been found that none are available). Only after that will it show the no accounts connected view
- Fixed the UI showing a serif font while the browser is still downloading Surt 🤢
- Fixed a number of useless console.warn/console.error messages which were making development just that little bit more annoying
- Reworded the transaction history deprecation notice
- Fixed `shortenAddress` to use an ellipsis `…` instead of three full stops `...`
- Removed the `theme.container.screens.2xl` size from the tailwind config (because we don't use it anywhere)
- Added eslint config to root of repo for vscode to find it
- Added missing polkadot.js overrides to root package.json (and upgraded them to a more recent version of pjs api)


yarn.lock
```diff
@@ -226,9 +226,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@azns/resolver-core@https://gitpkg.now.sh/api/pkg.tgz?url=TalismanSociety%2Fazero-id-resolver%2Fpackages%2Fresolver-core&commit=main":
+"@azns/resolver-core@https://gitpkg.vercel.app/api/pkg.tgz?url=TalismanSociety%2Fazero-id-resolver%2Fpackages%2Fresolver-core&commit=main":
```
chrisling-dev, line 229: Getting checksum dont meet error when I run yarn. Works fine after switching back to now.sh

eslint.config.mjs
```diff
@@ -0,0 +1,15 @@
+import recommended from '@talismn/eslint-config/react.js'
+import tseslint from 'typescript-eslint'
+
+export default tseslint.config(...recommended, {
+  rules: {
+    // TODO: turn this back on
+    '@typescript-eslint/no-explicit-any': 'off',
```
chrisling-dev, line 7: did you intend to turn this on within this PR?

apps/portal/src/util/format.ts
```diff
@@ -1,2 +1,2 @@
 export const shortenAddress = (address: string, padding = 4) =>
-  `${address.slice(0, padding)}...${address.slice(-padding)}`
+  `${address.slice(0, padding)}…${address.slice(-padding)}`
```
chrisling-dev, line 2: oo nice

apps/portal/src/routes/portfolio/index.tsx
```diff
@@ -1,13 +1,19 @@
...
 
 const routes = {
   element: (
-    <AccountConnectionGuard>
+    // If we suspend while waiting for the wallet, the underlying <Portfolio /> takes a
+    // significantly longer time to load due to the <Stakes /> component.
+    //
+    // TODO: Figure out how to fix the loading time, and then remove `noSuspense` so that we don't
+    // show a flash of the `Welcome to the Talisman Portal` UI when loading the page.
+    <AccountConnectionGuard noSuspense>
```
chrisling-dev, line 16: I'm not seeing any of the welcome page flashing, but the portfolio does get unbearably lag, not sure if this is a relevant concern

https://github.com/user-attachments/assets/b4b00cbc-8709-40d5-86c1-cd710f48e381

eslint.config.mjs
alecdwm: No! This is just a copy of what we have in `apps/portal/eslint.config.js`

chrisling-dev: ah ok np!

PR #1202, opened 2024-10-24
refactor/staking providers

UrbanWill:
# Description

Refactor staking providers table to allow sorting.

## Technical

- Every supported provider now has their own custom hook `useSomethingProviders`, where the basic provider information is fetched from state and formatted.
- Providers hooks data are combined in `useProvidersData`
- Combined data is passed down to `StakeProvidersTable`, where it is rendered. This is where the whole table UI gets rendered.
- Some cells required data, they have their own component to handle fetching and rendering. Every cell component follow the same structured.
- We need to keep track data of the fetched data to enable sorting, `useStakeValues` is used for that. 

## Type of change

- [x] Refactor
- [x] Ready to merge

### 🧪 **How to test:**
- Connect an account(s) with some staked tokens.
- Go to `/staking/positions`
- Click on the column names for sorting

Expected:
- Asc and desc sorting works for all columns
- UI matches previous table

### 🖼️ **Screenshots:**

https://github.com/user-attachments/assets/c14efb7a-6488-40e8-87ce-d718b734daf8


chidg (approved):
LGTM

PR #1209, opened 2024-11-15
feat: xcm improvements

alecdwm:
Includes several improvements/fixes for XCM:

- Use the same chaindata publish url as the `@talismn/chaindata-provider` library
- Update `@galacticcouncil/xcm-cfg` to support ForeignAssets
- Update `@talismn/balances` to support USDC on centrifuge
- Added empty state to the XCM form and updated form autofill code to match spec
- Added swaps-style info tab icons
- Improved token select button loading state
- Added token logo preloader for token select dialog
- Disabled MYTH mythos -> assethub route due to no isSufficient test (and therefore potential loss of funds)
- Improved min/max token amount messages
- Locked jotai and jotai-effect to one version to prevent a bug where the UI hangs and then eventually crashes ~5s after loading
- Fixed scrollable <Select> elements


chidg (approved):
LGTM, happy to merge this and then follow up with a few small fixes mentioned in Discord

PR #1214, opened 2024-11-29
fix: catch errors in useTotalRewards

UrbanWill:
# Description

Add try catch block to handle  ValidatorStakes useTotalRewrads errors without crashing.


chidg (approved):
LGTM!

PR #1216, opened 2024-12-02
chore: updated vMANTA copy to include gas fees instructions

UrbanWill:
# Description

- Added the following line to vMANTA staking side sheet:

`Make sure you have ETH on Manta Pacific in order to complete Manta liquid staking on Bifrost.`

### 🖼️ **Screenshots:**
Screenshot 2024-12-03 at 10 45 16


chidg (changes requested):
Can we put the new sentence into a different <p> element so there's some visual separation between this one and the previous?

chidg (approved):
LGTM!

PR #1217, opened 2024-12-02
fix: vDOT unbonding period 

UrbanWill:
# Description

Default unlock duration to 0 when formatting time distance.


chidg (approved):
LGTM

PR #1218, opened 2024-12-02
fix: more xcm improvements

alecdwm:
- Fixed DOT on AH errors (via xcm-sdk upgrade)
- Improved UI for large errors
- Improved UI for amount input
- Sort xcm token picker by fiat balance


chidg (approved):
LGTM

PR #1219, opened 2024-12-03
Refactor: wrap stake cells with Error Boundary

UrbanWill:
# Description

Wrapped stake providers and stake positions cells with ErrorBoundary. This change allows a single cell to throw an error without blocking the entire row.


apps/portal/src/components/widgets/staking/providers/components/StakeProvidersTable.tsx
```diff
@@ -66,6 +67,7 @@ const StakeProvidersTable = ({ dataQuery }: StakeProviderProps) => {
         header: 'Est. return',
         cell: ({ row }) => {
           return (
+            // <ErrorBoundary renderFallback={() => <>--</>}>
```
alecdwm, line 70: Is this one needed?

UrbanWill: Ops I goofed, should have removed this commented code. All cell errors are being handled in the .map, where the error boundary is then added.

alecdwm (approved):
LGTM 🚢

PR #1220, opened 2024-12-06
fix: some portfolio icon bugs

alecdwm:
Fixes:
- Don't add border-radius to chaindata icons
- Don't show chain name/id where only token is shown (and thus many chains are relevant)
- Alert triangle should be inline with balance


UrbanWill (approved):
🚢

PR #1223, opened 2024-12-09
refactor: better handle promises

UrbanWill:
# Description

- Do not mix async/await with .then
- Added try catch block
- Log error


chidg (approved):
LGTM!

PR #1244, opened 2025-01-03
Fix/ramp remove swap asset

UrbanWill:
# Description

Removed `swapAsset=DOT_DOT` from ramp URL. This was causing the Ramp to be "locked" to DOT on Polkadot only and did not allow the user to switch to a different asset.


alecdwm (approved):
<div align="center">

pacha-perfect

</div>

PR #1253, opened 2025-02-05
feat: support rootnet subnet staking

UrbanWill:
## Description

Support root net id for subnet staking.

### Technical 

Since changes are not in prod, a try catch block has been added to handle 'default' staking and the new subnet staking on root net with `netuid = 0`

### Balances fetching issue:

Could't find any docs on it, but FAFOing around I noticed the `getStakeInfoForColdkey` now expects an address str as arg when before it expected a address vector.

For anyone interested in checking themselves:

- Go to polkadot js on [bittensor testnet runtime calls](https://polkadot.js.org/apps/?rpc=wss://test.chain.opentensor.ai:443#/runtime)
- Select `stakeInfoRuntimeApi` and `getStakeInfoForColdkey`

Note:

For reasons beyond my comprehension, it was simply not possible to call the exact same methods without vectorizing the address. It will throw.

Eg:
```const response = (await api.rpc.state.call('StakeInfoRuntimeApi_get_stake_info_for_coldkey', address))```
instead of
``` const response = (await api.rpc.state.call('StakeInfoRuntimeApi_get_stake_info_for_coldkey', params)).toHex()```

Throws

### 🖼️ **Screenshots:**
Screenshot 2025-02-10 at 8 59 12

### DIFF
--- a/apps/portal/src/domains/staking/subtensor/atoms/accountStake.ts
+++ b/apps/portal/src/domains/staking/subtensor/atoms/accountStake.ts
@@ -45,7 +45,7 @@ export const accountStakeAtom = atomFamily(
             coldkey,
             hotkey,
             // make every stake a `bigint`, instead of a `number | bigint`, for consistency
-            stake: typeof stake === 'number' ? BigInt(stake) : stake,
+            stake: BigInt(stake),
           }))
           .filter(({ stake }) => stake !== 0n)
 
@@ -64,6 +64,7 @@ export const accountStakeAtom = atomFamily(
             coldkey,
             hotkey,
             netuid: BigInt(netuid),
+            // make every stake a `bigint`, instead of a `number | bigint`, for consistency
             stake: BigInt(stake),
           }))
           .filter(({ stake }) => stake !== 0n)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: apps/portal/src/domains/staking/subtensor/atoms/accountStake.ts [hunk 1]
```diff
@@ -35,10 +48,25 @@ export const accountStakeAtom = atomFamily(
...
-        return undefined
+        const response = (
+          await api.call['stakeInfoRuntimeApi']?.['getStakeInfoForColdkey']?.(address)
+        )?.toHuman() as DTaoStakeInfo[]
+
+        if (!Array.isArray(response)) return undefined
+
+        const stakes = response
+          // Filter out Subnet stakes for now
+          .filter(stake => Number(stake?.netuid) === ROOT_NETUID)
```
chidg, line 59: Why filter them out?
Even if just on testnet, it's still useful to see.

UrbanWill: This PR is for supporting the current functionality when dTAO goes live, which is rootnet only.

Displaying dTAO positions requires some extra logic and UI changes that are in the coming in the dTAO staking PR, where the user will be allowed to manage their positions for a given Validator and subnet

alecdwm (approved):
LGTM!

PR #1260, opened 2025-02-11
fix: add default values when api fails

UrbanWill:
## Description

Add default values when Taostats api calls fails.


apps/portal/src/components/widgets/staking/subtensor/StakeForm.tsx
```diff
@@ -91,8 +91,7 @@ export const IncompleteSelectionStakeForm = (props: IncompleteSelectionStakeForm
 
 const EstimatedRewards = (props: { amount: bigint; delegateHotkey: string }) => {
   const tokenAmount = useRecoilValue(useNativeTokenAmountState())
-  // const apr = useApr()
```
alecdwm, line 94: I think this snuck in

UrbanWill: That was deliberate, I removed the commented code that shouldn't be there :)

alecdwm: OH lol, caffeine-deficient moment on my part 😭

alecdwm (approved):
:shipit:

PR #1266, opened 2025-02-20
Feat/update d tao balances

UrbanWill:
# Description

- Added missing key to StakeInfo struct
- Refactor fetching staking data to use react-query and pooling method for up to positions & prices
- Invalidate cache on staking/unstaking actions, display fresh data

### 🧪 **How to test:**
- Stake/unstake dTao (or Tao).
- Balance should be reflected as soon as the transaction is confirmed
- Worst case scenario it should be pooled every 30s

### 🖼️ **Screenshots:**

https://github.com/user-attachments/assets/b209cda4-0491-4e90-a70c-131e608c6a55


alecdwm (approved):
:shipit:

PR #1270, opened 2025-03-04
feat: store combined data in state, return if api call fails

UrbanWill:
# Description

Store combined api response data in state, do not update state if api call fails.

## Technical

The API calls are being pre populated with placeholder data, which then gets hydrated with fresh data from api responses. The issue arises when the api response fail, and placeholder data then gets replaced by `null`


UrbanWill:
Note to self: https://github.com/TalismanSociety/talisman-web/issues/1271

0xKheops (approved):

PR #1285, opened 2025-03-26
Feat/bittensor move stake

UrbanWill:
# Description

Enable users to switch staked position's validators without charging fees.

## Technical 

Create new `useMoveStake` hook to handle moving staked positions. Note: we should move `useAddStakeForm` and `useUnstakeForm` hooks that are currently in the form file, for better maintenance.

### 🖼️ **Screenshots:**

https://github.com/user-attachments/assets/489e33d1-912c-49fb-a0ca-d98be9a0b080


alecdwm (approved):
LGTM!

PR #1301, opened 2025-04-16
chore: change default sort order

UrbanWill:
# Description

- Change subnet select modal default sort order to Total Alpha
- Fix Token units wrapping into two lines, display "29k TAO" instead of "29,000.00 TAO" 

### 🖼️ **Screenshots:**

Screenshot 2025-04-16 at 16 18 45


UrbanWill:
> Member

Can you expend on that?

We are already sorting validators by total Tao staked, are you suggesting to fetch and display the total alpha staked through a given validator, and then default the sorting order of those validators by total alpha? I think we could to that, but it would also require to block validator selection until a subnet is selected

alecdwm (approved):
:shipit:

PR #1329, opened 2025-05-24
feat: switch to broker as a service for CF

CumpsD:
* Added the Talisman **Public** API key, this can be exposed.
* Switched to use it for mainnet, testnet is just the common testnet key.
* Removed `boostFeeBps` on quote, this does not exist

Remaining TODO's to make the integration work again:
* `fillOrKillParams` is now a required field.


alecdwm:
Tyvm @CumpsD! Merging this into a local branch so I can push commits and add the `fillOrKillParams` prop

PR #1358, opened 2025-07-09
chore: comment out total staked while the value is incorrect

UrbanWill:
# Description

Comment out Total staked amount on Staking dashboard, not all assets are included in the calculation and the amount is inccorrect.

Related to: https://github.com/TalismanSociety/talisman-web/issues/1309, but more assets are may be left out of the total calc.


alecdwm (approved):
:shipit:

